PR #415, opened 2013-12-18
Make the ring class injectable

zmack:
This adds the ability to pass a class into the options hash to be used as a ring.

The rationale behind it is that there are many other memcache libs using different hash ring logic, and being able to enforce that logic consistently throughout your applications shouldn't involve actually hacking dalli.

If you guys feel this would be a useful addition, I'll most likely add some documentation to this, just to be sure the use case for this is clear.


orend:
@mperham is the ring algorithm you're using for selecting a server to write to compatible with java-memcached-client's?

petergoldstein:
Given that this PR didn't get a lot of attention, and never quite made it over the line, I'm going to close it.  I think an injectable algorithm is a good idea, but I think it probably makes more sense to start from scratch at this point.

Issue #251, opened 2012-08-12
Silently doesn't work if TTL is over 30 days


henrik:
This works:

```
require "dalli"
dc = Dalli::Client.new('localhost:11211')
dc.set "test", "works", 60*60*24*30
puts(dc.get("test") || "fails")
```

This fails:

```
require "dalli"
dc = Dalli::Client.new('localhost:11211')
dc.set "test", "works", 60*60*24*30 + 1
puts(dc.get("test") || "fails")
```

This was confusing to me and took a while to debug. I now know that memcache expiry given in seconds can't be higher than 30 days (I should probably just use 0 = never expire), but perhaps Dalli could raise an exception for expiry times over 30 days to help you catch these mistakes?

If you'd like, I'll make a pull request, but I wanted to check if you want this behavior, first.

Issue #55, opened 2010-12-31
session_store silently fails with :expire_after => 120.days


ejw:
I have not done a lot of testing on this, but it took me a while to track down my particular issue. If I set the following:

Rails.application.config.session_store :dalli_store, :memcache_server => '127.0.0.1', :key => '_zedkit', :expire_after => 120.days

The Dalli session store will silently not work. No sessions get saved, Rails does not seem to even update the session with the csrf_token. If I set it to:

  Rails.application.config.session_store :dalli_store, :memcache_server => '127.0.0.1', :key => '_zedkit', :expire_after => 30.days

Sessions work as expected.

In your examples all the expire_after settings are quite low, but in my case, I keep the session around for a long time to record the user login, for example, and other interface information to improve usability. I don't want the session expiring unless the user really has gone away and is not likely to come back. I would like to set an expiry much larger than 30.days even.

I have a memcache cluster that maintains the "persistence" that is good enough for these purposes. Anyway, setting a much higher expiry should work.

Of course, the issue is not likely 120.days exactly, but some length of time over 30.days that causes the issue.

PR #436, opened 2014-02-26
Fix for expiration intervals longer than 30 days

leonid-shevtsov:
I have found out (the hard way) that memcached silently doesn't store values with expiration over 30 days. (previously: #55 #251 #357)

Upon studying the code, I've decided that the best way to handle this is to transparently rewrite long expiration intervals into expiration timestamps, as the Memcached API asks.

Tests are provided, I've also tested the code integrated into my own environment, a debug message is logged when the translation happens, and nothing changes in the documented API.

What will change in existing apps is
- apps that unknowingly use bad expiration intervals (such as 1 year) will suddenly start caching as intended
- apps that knowingly exploit the "bug" will get wrong expiration timestamps in the far future.

I think compliance to APIs is worth breaking the latter. An safer alternative could be to provide both `:expires_in` and `:expires_on` options, and throw exceptions on invalid timestamps and intervals, but Rails, for one, doesn't even have `:expires_on`.

### DIFF
--- a/History.md
+++ b/History.md
@@ -7,6 +7,7 @@ HEAD
 - Rack session will check if servers are up on initialization (arthurnn, #423)
 - Add support for IPv6 addresses in hex form, ie: "[::1]:11211" (dplummer, #428)
 - Add symbol support for namespace (jingkai #431)
+- Support expiration intervals longer than 30 days (leonid-shevtsov #436)
 
 2.7.0
 ==========


annaswims:
For anyone else that stumbles across this problem, Rails adds 5 minutes to the ttl for reasons that I don't understand.  So, `:expires_on => 30.days `  ends up being being greater than 30 days https://github.com/rails/rails/commit/13d8777b63f93426f6a1101f066cc3cf218109d3

PR #481, opened 2014-08-13
Make sure LocalCache respects Entry unwrapping and :raw option

sj26:
Trying to set a counter value and using increment while inside a LocalCache block currently fails. LocalCache expects `read_entry`,  `write_entry` and  `delete_entry` to be dealing with `Entry` instances, but Dalli chooses to deal with unwrapped values. This means that writes for raw values would be subsequently read as a raw then `Entry` wrapped object.

This addresses the failure by making sure the `read_entry` interface unwraps potential Entry instances from LocalStore and performs a `to_s` on the unwrapped value if raw is requested. The implementation is similar to the [workaround in activesupport for MemCacheStore's raw support](https://github.com/rails/rails/blob/82e28492e7c581cdeea904464a18eb11118f4ac0/activesupport/lib/active_support/cache/mem_cache_store.rb#L175-L195).

The `write_entry` and `delete_entry` interfaces seem to work sufficiently without this workaround, although there might be other undiscovered complications. It might be possible that `write_entry` is unpredictable and LocalStore may be receiving some Entry-wrapped values and other raw values. Further testing is probably necessary, but this pull should address the main issue.

### DIFF
--- a/lib/active_support/cache/dalli_store.rb
+++ b/lib/active_support/cache/dalli_store.rb
@@ -68,6 +68,7 @@ module ActiveSupport
         end
 
         extend Strategy::LocalCache
+        extend LocalCacheEntryUnwrapAndRaw
       end
 
       ##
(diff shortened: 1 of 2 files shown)


sj26:
@mperham any movement on this?

sj26:
@mperham @petergoldstein any movement on this? we still experience this problem.

petergoldstein:
@sj26 Can you please rebase this against mperham/dalli:master?  Then I'll take a look.  Thanks.

sj26:
@petergoldstein rebased. thanks!

sj26:
We're currently using https://gist.github.com/sj26/6c9ddc1740e40b95c0db in production

PR #492, opened 2014-10-12
Add namespace as local option

### LINKED ISSUE #490: namespace_key not available
I don't know if I am doing something wrong but I can't use a namespace_key.

The rails active support cache allow us to define it like that:

```
<% cache(object, namespace: 'namespace') %>
```

See here for the definition:
https://github.com/rails/rails/blob/master/activesupport/lib/active_support/cache.rb#L-534

It seems that it is not working when using dalli.
Is it because because it was not in the plan to implement this feature?

alaibe:
fixes #490 
- add namespaced_key method from active support
- add new test to ensure you can use a namespace locally

I am just not sure about read and write multi.


tpitale:
Any idea when this might make it into a gem release?

PR #514, opened 2015-02-23
Fix instrumentation of hits for falsey values

AlexRiedler:
Handle case when the entry is the boolean value 'false' that it should still be considered a hit.

- I did this as a quick inline fix, will still need to do verification.


petergoldstein:
@AlexRiedler I'm inclined to merge this, although I'd like a spec before I do.  This is a pretty subtle (and hence easy to regress) bug.  Could you do a quick spec?

AlexRiedler:
@petergoldstein that was more annoying that I thought it would be :( but there is some tests now!

PR #531, opened 2015-04-24
Add ability to use Proc as 'expires_in' key

avokhmin:
Sometimes calculation of `expires_in` can spend a lot of time, so, will be great to do it only if it needed.

Example:

```
[19] pry(main)> x = Rails.cache.fetch('test', expires_in: -> { (1..5).each{ |i| puts(i); sleep(1) }; 10  }) { 'hello' }
Cache read: test ({:expires_in=>#<Proc:0x0000010f058d08@(pry):9 (lambda)>})
Cache generate: test ({:expires_in=>#<Proc:0x0000010f058d08@(pry):9 (lambda)>})
Cache write: test ({:expires_in=>#<Proc:0x0000010f058d08@(pry):9 (lambda)>})
1
2
3
4
5
=> "hello"
[20] pry(main)> x = Rails.cache.fetch('test', expires_in: -> { (1..5).each{ |i| puts(i); sleep(1) }; 10  }) { 'hello' }
Cache read: test ({:expires_in=>#<Proc:0x0000010f1b8f90@(pry):10 (lambda)>})
Cache fetch_hit: test ({:expires_in=>#<Proc:0x0000010f1b8f90@(pry):10 (lambda)>})
=> "hello"
```

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,11 @@
 Dalli Changelog
 =====================
 
+2.7.5
+==========
+
+- Add ability to use `Proc` as `:expires_in` key (avokhmin, #531)
+
 2.7.4
 ==========
 
--- a/lib/dalli/cas/client.rb
+++ b/lib/dalli/cas/client.rb
@@ -35,7 +35,7 @@ module Dalli
     # Set the key-value pair, verifying existing CAS.
     # Returns the resulting CAS value if succeeded, and falsy otherwise.
     def set_cas(key, value, cas, ttl=nil, options=nil)
-      ttl ||= @options[:expires_in].to_i
+      ttl ||= expires_in_value @options
       perform(:set, key, value, ttl, cas, options)
     end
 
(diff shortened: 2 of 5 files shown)


be9:
:+1:

k41n:
:+1:

avokhmin:
Not sure what happens on travis-ci, but locally oall tests passes for `Rails 3.2.21` and `jruby 1.7.19`, see [log](https://gist.github.com/avokhmin/f2b500b7761d0387d280)

petergoldstein:
If I'm understanding the goal of this code correctly, then I'm inclined not to merge it.  Essentially this PR is  only useful in the Rails cache case.  In the case where the Dalli::Client is being accessed directly, the TTL can be passed in explicitly, so none of this complexity is required to control the cost of the TTL calculation.  It's only in the case of Rails Cache that the TTL is controlled exclusively through the expires_in initialization option.

In this case, I'll defer to Rails to determine the allowed types for the expires_in option.  Looking at the docs, there's no mention of using a Proc as you describe.  See here - http://guides.rubyonrails.org/caching_with_rails.html

If you really want this, I'd attempt to get a PR into Rails.

Issue #400, opened 2013-10-29
"No server available" error on bad parameter


tjchambers:
I was trying to understand how some of my code was getting a No server available when it was available. I was able to isolate the problem to a bad parameter to incr method. 

The code that creates this error is https://gist.github.com/tjchambers/7213452 and the
results from running the code are: https://gist.github.com/tjchambers/7213479 .

I do not receive the internal error that is reported when I run my code, just the Dalli::RingError: No server available which is obviously not the correct error for the condition.

jstejada:
:+1: any updates on this? I'm having the same issue, getting a RingError, but my memcached server is definitely up. Still haven't figured out what the real cause is, but I think it is because of this

inspire22:
Previously, sending an invalid key (i.e. a space or quote in it) would sometimes affect this and/or show an odd downtime message - not sure if that's still an issue or not.  Also check your memcached version...

petergoldstein:
This should be fixed by #582

PR #547, opened 2015-07-29
Validate ttl class

eagletmt:
When the application gives an invalid ttl value, dalli considers it to
be a server error and detaches the server, which leads to "No server
available" error. The ttl value should be validated and raise an
appropriate error before trying to perform caching.

related to: #400

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -85,6 +85,7 @@ module Dalli
     # written to the cache and returned.
     def fetch(key, ttl=nil, options=nil)
       ttl ||= @options[:expires_in].to_i
+      validate_ttl(ttl)
       options = options.nil? ? CACHE_NILS : options.merge(CACHE_NILS) if @options[:cache_nils]
       val = get(key, options)
       not_found = @options[:cache_nils] ?
@@ -110,6 +111,7 @@ module Dalli
     # - true if the value was successfully updated.
     def cas(key, ttl=nil, options=nil)
       ttl ||= @options[:expires_in].to_i
+      validate_ttl(ttl)
       (value, cas) = perform(:cas, key)
       value = (!value || value == 'Not found') ? nil : value
       if value
@@ -120,6 +122,7 @@ module Dalli
 
     def set(key, value, ttl=nil, options=nil)
       ttl ||= @options[:expires_in].to_i
+      validate_ttl(ttl)
       perform(:set, key, value, ttl, 0, options)
     end
 
@@ -128,6 +131,7 @@ module Dalli
     # on the server.  Returns truthy if the operation succeeded.
     def add(key, value, ttl=nil, options=nil)
       ttl ||= @options[:expires_in].to_i
+      validate_ttl(ttl)
       perform(:add, key, value, ttl, options)
     end
 
@@ -136,6 +140,7 @@ module Dalli
     # on the server.  Returns truthy if the operation succeeded.
     def replace(key, value, ttl=nil, options=nil)
       ttl ||= @options[:expires_in].to_i
+      validate_ttl(ttl)
       perform(:replace, key, value, ttl, 0, options)
     end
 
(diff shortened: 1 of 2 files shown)


eagletmt:
When non-integer ttl is given, ArgumentError is raised from `Dalli::Server#sanitize_ttl` .
https://github.com/mperham/dalli/blob/v2.7.4/lib/dalli/server.rb#L474

```ruby
dalli = Dalli::Client.new('127.0.0.1:11211')
dalli.set(Time.now.to_s, 'bar', Time.now)
```

```
E, [2015-07-30T04:20:56.297442 #28964] ERROR -- : Unexpected exception during Dalli request: ArgumentError: comparison of Time with 2592000 failed
E, [2015-07-30T04:20:56.297521 #28964] ERROR -- : /home/eagletmt/.ghq/github.com/mperham/dalli/lib/dalli/server.rb:479:in `>'
	/home/eagletmt/.ghq/github.com/mperham/dalli/lib/dalli/server.rb:479:in `sanitize_ttl'
	/home/eagletmt/.ghq/github.com/mperham/dalli/lib/dalli/server.rb:279:in `set'
	/home/eagletmt/.ghq/github.com/mperham/dalli/lib/dalli/server.rb:67:in `request'
	/home/eagletmt/.ghq/github.com/mperham/dalli/lib/dalli/options.rb:18:in `block in request'
	/usr/lib/ruby/2.3.0/monitor.rb:213:in `mon_synchronize'
	/home/eagletmt/.ghq/github.com/mperham/dalli/lib/dalli/options.rb:17:in `request'
	/home/eagletmt/.ghq/github.com/mperham/dalli/lib/dalli/client.rb:333:in `perform'
	/home/eagletmt/.ghq/github.com/mperham/dalli/lib/dalli/client.rb:107:in `set'
	/home/eagletmt/.ghq/github.com/mperham/dalli/a.rb:7:in `<main>'
W, [2015-07-30T04:20:56.297589 #28964]  WARN -- : 127.0.0.1:11211 is down
```

petergoldstein:
@eagletmt I'm a little confused as to what problem this is addressing.  When I look at the existing code, it appears that the introduction of sanitize_ttl already addresses the issue in #400, at least as far as bad ttl parameters go.  Can you provide a little more detail?  Thanks.

eagletmt:
`sanitize_ttl` raises ArgumentError when `ttl` isn't comparable with `MAX_ACCEPTABLE_EXPIRATION_INTERVAL` .
https://github.com/petergoldstein/dalli/blob/v2.7.4/lib/dalli/server.rb#L473-L480
A example code is already listed in my comment https://github.com/petergoldstein/dalli/pull/547#issuecomment-126066933 .

The programming error results in "No server available" error, which is very critical because all caching mechanism become unavailable.
This patch minimizes the effect of the programming error (i.e., passing an invalid ttl value) by raising a correct error and protecting from "No server available" error.

petergoldstein:
@eagletmt  I don't see that the current behavior results in a "No server available" error.  The bug you reference ( #400  ) was filed before `sanitize_ttl` was added, and I'm not observing the behavior you're reporting.  Your test case does nothing to show that the behavior you're reporting exists in the current codebase

eagletmt:
I posted a runnable code here https://gist.github.com/eagletmt/31b5bad78238b64279bb .
Running `ruby pr547.rb > pr547.stdout.log 2> pr547.stderr.log` should reproduce the problem I mentioned.

petergoldstein:
@eagletmt Ok, thanks for clarifying with a test case.  Now I'm seeing the issue.

That said, I don't think this PR is the right solution.  It explicitly precludes using ActiveSupport::Duration as a ttl (which is likely to be overly restrictive in a Rails context), to resolve what is essentially a corner case due to caller error.  I think we should allow any object which implements `to_i` as a ttl value, even if in some cases (like a string value) this may allow inadvertent caller errors.  That would match the behavior for the expires_in option

I do agree that such a caller error should not be escalated to a `Dalli::RingError`, and it needs to be handled in the `Dalli:Client`.  I've put together a PR that is less restrictive on ttl values in line with my comments above and should resolve the issue - #582

Thanks again for raising the issue.

eagletmt:
:ok_woman: 
Thank you for your better fix.

PR #549, opened 2015-08-11
Support touch in dalli_store

hirocaster:
Support touch command in dalli_store

### DIFF
--- a/lib/active_support/cache/dalli_store.rb
+++ b/lib/active_support/cache/dalli_store.rb
@@ -136,6 +136,16 @@ module ActiveSupport
         end
       end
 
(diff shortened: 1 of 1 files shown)


petergoldstein:
@hirocaster I have basically the same concern about this PR as I did about #550.  Why can't you just use the `dalli` method to access the underlying `Dalli::Client` to do the `touch`?  As far as I'm concerned, `ActiveSupport::Cache::Dalli` is a Rails adapter class, and its methods should be driven by Rails

Thoughts?

hirocaster:
Thank you feedback.
I will use #dalli (Sorry, I was not aware of this method.)

PR #554, opened 2015-09-10
Escape namespace when using a Regexp to replace it

speckins:
I experienced read_multi results coming back with nil keys.  Eventually I tracked it down to the namespace, which contained a base64-encoded component, but only when it contained the "+" character.

Normal

    irb(main):009:0> Rails.cache.options[:namespace] = 'abc'
    => "abc"
    irb(main):010:0> Rails.cache.write 'foo', 'bar'; puts Rails.cache.read 'foo'; Rails.cache.read_multi 'foo'
    bar
    => {"foo"=>"bar"}

With special characters ("+")

    irb(main):011:0> Rails.cache.options[:namespace] = 'a+c'
    => "a+c"
    irb(main):012:0> Rails.cache.write 'foo', 'bar'; puts Rails.cache.read 'foo'; Rails.cache.read_multi 'foo'
    bar
    => {nil=>"bar"}

Of course, other special regex characters will cause this, too, or possibly raise RegexpErrors (e.g., "[]").  Properly escaping the namespace when using a regexp to remove it from the key solves this issue.

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -371,7 +371,7 @@ module Dalli
     end
 
     def key_without_namespace(key)
-      (ns = namespace) ? key.sub(%r(\A#{ns}:), '') : key
+      (ns = namespace) ? key.sub(%r(\A#{Regexp.escape ns}:), '') : key
     end
 
     def namespace
--- a/test/test_active_support.rb
+++ b/test/test_active_support.rb
@@ -204,6 +204,19 @@ describe 'ActiveSupport' do
       end
     end
 
+    it 'support read_multi with special Regexp characters in namespace' do
+      with_activesupport do
+        memcached_persistent(@port) do
+          # /(?!)/ is a contradictory PCRE and should never be able to match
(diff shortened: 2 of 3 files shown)


petergoldstein:
@speckins Can you please create a spec to demonstrate the problem and provide protection against regression?   Thanks.

speckins:
@petergoldstein Sure.  I almost did with the original PR but didn't think it significant enough.

Here's what they look like when they fail:

    1) Failure:
    Dalli#test_0001_handle special Regexp characters in namespace with get_multi:
    Expected: {"a"=>1, "b"=>2}
      Actual: {"(?!):a"=>1, "(?!):b"=>2}

    2) Failure:
    Dalli#test_0002_support read_multi with special Regexp characters in namespace:
    Expected: {"abc"=>123, "xyz"=>456}
      Actual: {nil=>456}

The `Regexp.escape` addition causes both to pass.  (If you sign off of both of these, I will squash them into the previous commit and `push -f`.)

speckins:
I squashed the specs and rebased to dalli:master.  Is there anything holding up this PR?

petergoldstein:
Looks good, thanks!

Issue #208, opened 2012-05-05
Detect and handle sockets shared by processes


mperham:
See redis/redis-rb#226

With this, Dalli should no longer need to be reset in Unicorn and Passenger's after_fork block.

Issue #504, opened 2014-12-03
When package the newest code to rubygems?


yuc-zhu:
The 2.7.2 has bug when use sasl, when you package the new code to rubygems?

petergoldstein:
I just pushed 2.7.3

PR #556, opened 2015-09-10
Do not treat socket sharing as a failure

### LINKED ISSUE #440: Hi, I am getting this issu,RuntimeError: Cannot share client between multiple processes
config file:
defaults:
  ttl: 1800
  readonly: false
  urlencode: false
  c_threshold: 10000
  compression: true
  debug: false
  namespace: ""
  sessions: true
  fragments: true
  memory: 128
  servers:
      - my-test-server:11211
  benchmarking: true
  raise_errors: true
  fast_hash: false
  fastest_hash: false

error:
 my-test-server:11211 failed (count: 0) RuntimeError: Cannot share client between multiple processes
# <Dalli::NetworkError: Socket operation failed, retrying...>

retrying request with new server
Dalli::Server#connect my-test-server:11211

I have two nodes of AWS elastic cache server, when multiple users hit the same cache.

Am i missing some configs, please help.
Thanks in advance.

wjordan:
Fixes #440.

Since #208 "Dalli no longer needs to be reset after Unicorn/Passenger fork," which implies that this situation will be transparently handled by the library without logging any errors, without any workarounds required. However, the confusion started with #368 when an `info` log message was added that implied that this situation was not permitted and was causing a `RuntimeError` on the server.

Since sharing a socket with client processes is now handled perfectly fine by the library, this PR changes the log message caused by this socket-sharing condition from a scary:

```
my-test-server:11211 failed (count: 0) RuntimeError: Cannot share client between multiple processes
```

to a more reasonable:

```
Fork detected, re-connecting child process...
```

It also skips incrementing the `@fail_count` on this condition, so the full count of `:socket_max_failures` are available to the child process.

### DIFF
--- a/lib/dalli/server.rb
+++ b/lib/dalli/server.rb
@@ -207,7 +207,17 @@ module Dalli
 
     def verify_state
       failure!(RuntimeError.new('Already writing to socket')) if @inprogress
-      failure!(RuntimeError.new('Cannot share client between multiple processes')) if @pid && @pid != Process.pid
+      if @pid && @pid != Process.pid
+        message = 'Fork detected, re-connecting child process...'
+        Dalli.logger.info { message }
+        reconnect! message
+      end
+    end
+
+    def reconnect!(message)
+      close
+      sleep(options[:socket_failure_delay]) if options[:socket_failure_delay]
+      raise Dalli::NetworkError, message
     end
 
     def failure!(exception)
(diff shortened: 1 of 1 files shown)


wjordan:
the Travis CI build failure [#504.3](https://travis-ci.org/mperham/dalli/jobs/79745275) on this PR seems to be the result of some unrelated test flakiness (compare to [#491.3](https://travis-ci.org/mperham/dalli/jobs/70187868))

PR #559, opened 2015-09-14
support the fetch operation with nil values

### LINKED ISSUE #294: Inconsistent fetch nil behavior
Example code, run in rails console

```ruby
#
# expected result, prints "fetching" one time
#
cache.flush
key = "t_nil";
p "value from get = #{cache.get(key).inspect}";
r = cache.fetch(key){ puts "fetching #{key}"; 5 };
r = cache.fetch(key){ puts "fetching #{key}"; 3 };
r = cache.fetch(key){ puts "fetching #{key}"; 3 };
r = cache.fetch(key){ puts "fetching #{key}"; 3 };
r = cache.fetch(key){ puts "fetching #{key}"; 3 };
p "value from fetch = #{r.inspect}";
p "value from get = #{cache.get(key).inspect}";
""

#
# unexpected result, prints "fetching" five times
#
cache.flush
key = "t_nil";
p "value from get = #{cache.get(key).inspect}";
r = cache.fetch(key){ puts "fetching #{key}"; nil };
r = cache.fetch(key){ puts "fetching #{key}"; 3 };
r = cache.fetch(key){ puts "fetching #{key}"; 3 };
r = cache.fetch(key){ puts "fetching #{key}"; 3 };
r = cache.fetch(key){ puts "fetching #{key}"; 3 };
p "value from fetch = #{r.inspect}";
p "value from get = #{cache.get(key).inspect}";
""
```

Two issues here:

1) In dalli client, nil and missing value are not distinguished from each other, so if a fetch block returns nil, calling the same fetch block will result in the block being run once more on the next call. If the philosophy is "you should not cache nil, it represents no value" then that may be working as intended. I personally disagree with this approach and find it annoying because I like to cache expensive method calls where nil is a perfectly acceptable return value.

2) If nil gets cached on the first fetch, all subsequent fetch blocks will be run, even if they return a non-nil value, and that non-nil value will not write over the nil value. The API seems to be making nil considerably more special than a missing cached value, so I am guessing that is a bug.

Thoughts?

wjordan:
Fixes #294.

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -27,6 +27,7 @@ module Dalli
     # - :compress - defaults to false, if true Dalli will compress values larger than 1024 bytes before sending them to memcached.
     # - :serializer - defaults to Marshal
     # - :compressor - defaults to zlib
+    # - :cache_nils - defaults to false, if true Dalli will not treat cached nil values as 'not found' for #fetch operations.
     #
     def initialize(servers=nil, options={})
       @servers = normalize_servers(servers || ENV["MEMCACHE_SERVERS"] || '127.0.0.1:11211')
@@ -52,8 +53,9 @@ module Dalli
 
     ##
     # Get the value associated with the key.
+    # If a value is not found, then +nil+ is returned.
     def get(key, options=nil)
-      perform(:get, key)
+      perform(:get, key, options)
     end
 
     ##
@@ -71,10 +73,24 @@ module Dalli
       end
     end
 
+    CACHE_NILS = {cache_nils: true}.freeze
+
+    # Fetch the value associated with the key.
+    # If a value is found, then it is returned.
+    #
+    # If a value is not found and no block is given, then nil is returned.
+    #
+    # If a value is not found (or if the found value is nil and :cache_nils is false)
+    # and a block is given, the block will be invoked and its return value
+    # written to the cache and returned.
     def fetch(key, ttl=nil, options=nil)
       ttl ||= @options[:expires_in].to_i
+      options = options.nil? ? CACHE_NILS : options.merge(CACHE_NILS) if @options[:cache_nils]
       val = get(key, options)
-      if val.nil? && block_given?
+      not_found = @options[:cache_nils] ?
+        val == Dalli::Server::NOT_FOUND :
+        val.nil?
+      if not_found && block_given?
         val = yield
         add(key, val, ttl, options)
       end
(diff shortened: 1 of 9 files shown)


petergoldstein:
@wjordan Thanks for the PR.

A couple of quick comments:

1. Can we rename the option from 'allow_nil' to 'cache_nils'?  I think that expresses the intention better.
2. Please add some documentation for the option as part of the PR
3. It looks like with this PR the new behavior is now the default behavior for fetch.  I'm concerned about changing that default behavior, as users of the library may depend on the earlier behavior.  Can we change the code so that the new behavior is opt-in, and we can discuss changing the default as part of the next major release?
4. Can you please add a comment - e.g. "Implements Null Object Pattern for caching nils" before the NilObject class definition?  As is, it's just sitting in the middle of the file, and a casual reader will likely be unclear as to the purpose of the class.

lib/dalli/client.rb
```diff
@@ -72,9 +72,10 @@ def get_multi(*keys)
     end
 
     def fetch(key, ttl=nil, options=nil)
+      options ||= {}
```
Napolskih: good request!

we can do this more optimally:
```ruby
  DEFAULT_FETCH_OPTIONS = {allow_nil: true}.freeze

  def fetch(key, ttl=nil, options=nil)
    options ||= DEFAULT_FETCH_OPTIONS
    ttl ||= @options[:expires_in].to_i
    val = get(key, options)
    ...
  end
```

wjordan:
Thanks for taking a look @petergoldstein, I'll look into making the suggested changes.

I've also just added another commit (actually two commits, the first commit refactors the `fetch` part of the `active_support` test suite into individual tests) for `DalliStore#fetch` to also support nil values. This method calls `Dalli::Client#read` (through #read_entry) rather than `Dalli::Client#fetch` directly, though it should support the same nil-value behavior. Please let me know if these extra changes look good as well.

wjordan:
@petergoldstein updated based on your feedback, please take a look- I could rebase into a single commit if that's preferred.

I also incorporated @Napolskih's suggestion into `client.rb` [here](https://github.com/mperham/dalli/pull/559/files#diff-529a991515d2e5a278fa5b2cb3e9a83eR88), though it would help to get some feedback on this particular syntax - I'm not sure I know the most efficient way to keep this inner-loop code as heavily-optimized as possible to ensure no general performance regressions (however slight) result from this feature.

petergoldstein:
@wjordan If you could rebase and squash, I'll take a look at the resulting PR.  Thanks for incorporating my feedback.

wjordan:
@petergoldstein rebased and squashed.

petergoldstein:
This looks good.  I'll merge it in.

Issue #286, opened 2012-11-11
Support stale cache reads


mperham:
We have a case where we call a third party service which goes down a lot.  I would like to cache the result and serve it if the value cannot be fetched, eg:

``` ruby
dalli.fetch('foo', :stale_after => 1.hour, :expires_in => 12.hours, :error_handler => nil) do
  call_service
end
```

I'm inclined to make this a new method and not add more logic to `fetch`.

PR #560, opened 2015-09-18
Add support for `:race_condition_ttl` option

wjordan:
This PR adds support for the `:race_condition_ttl` option from `ActiveSupport::Cache`. This implementation uses an `Entry` wrapper object to append the secondary expiration to the value being stored.

An alternative approach avoids using a wrapper object by writing/reading an extra metadata-only key with the secondary expiration (as attempted in #378). However, this approach performed worse in my local tests, likely due to the need to send multiple `get` requests to the server on each `#fetch` (`multi_get` was even slower for some reason).

To avoid performance regressions, the wrapper object is only used when the `:race_condition_ttl` option is explicitly specified in a `#write` or `#fetch` call. A single instance object is reused to reduce garbage creation, and custom serialization/deserialization methods limit the total per-object overhead to only 8 bytes (storing the `expires_at` timestamp as a double-precision float).

The included tests are adapted from the ActiveSupport tests, however the test objects are expired via `sleep` commands instead of stubbing `Time.now` as done in `ActiveSupport::Cache`\- if it's possible to stub Dalli to manually-expire objects on the server, that would help here.

This should support the `:stale_after` use-case described in #286 (see the `is safe` test in particular), but I'm not 100% sure about this.

### DIFF
--- a/lib/active_support/cache/dalli_store.rb
+++ b/lib/active_support/cache/dalli_store.rb
@@ -98,14 +98,7 @@ module ActiveSupport
         if block_given?
           entry = not_found
           unless options[:force]
-            entry = instrument(:read, namespaced_name, options) do |payload|
-              read_entry(namespaced_name, options).tap do |result|
-                if payload
-                  payload[:super_operation] = :fetch
-                  payload[:hit] = result != not_found
-                end
-              end
-            end
+            entry = find_cached_entry(namespaced_name, options)
           end
 
           if entry == not_found
@@ -129,6 +122,7 @@ module ActiveSupport
 
         instrument(:read, name, options) do |payload|
           entry = read_entry(name, options)
+          entry = entry.is_a?(Entry) ? entry.value : entry
           payload[:hit] = !entry.nil? if payload
           entry
         end
@@ -141,7 +135,17 @@ module ActiveSupport
         instrument(:write, name, options) do |payload|
           with do |connection|
             options = options.merge(:connection => connection)
-            write_entry(name, value, options)
+            # Support :race_condition_ttl by appending a secondary expiration to the value.
+            if options[:race_condition_ttl] &&
+              (ttl = options[:expires_in] || @options[:expires_in])
+              options[:expires_in] = ttl + options[:race_condition_ttl] * 2
+              entry = Entry.new
+              entry.value = value
+              entry.expires_at = ttl + Time.now.to_f
+              write_entry(name, entry, options)
+            else
+              write_entry(name, value, options)
+            end
           end
         end
       end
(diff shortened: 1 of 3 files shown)


petergoldstein:
@wjordan At first glance this doesn't look thread safe to me.  What prevents multiple threads from simultaneously accessing and modifying the shared ENTRY variable?  Is there a mutex higher in the call stack?

wjordan:
You're right, I didn't consider the multithreaded scenario for that reused entry object. I'll update the branch to use a thread-local variable soon.

petergoldstein:
@wjordan Honestly, that all feels like premature optimization.  I'd just create an object per call and run some comparison tests.  See whether there's an actual performance problem to be fixed before substantially increasing the complexity of the code.

Also, can you please rebase now that your other PR was merged?  Thanks.

wjordan:
Fair enough- the benchmark was only about 4-5% faster with the optimization (with a margin of error around ~10%), so the garbage-optimization was not necessary. I've added a commit that removes this (and simplifies some of the marshal-load logic as well), and will rebase the branch soon.

wjordan:
Rebased with a couple more changes/tweaks.

- I made the per-object serialization [overhead](https://github.com/wjordan/dalli/commit/f1a2ab8f10f5c8abe105c48bf5c1871ee942a2d6#diff-90586e1eaac5757610f69a1c1ae4c45eR364) explicit (the number is verified through a unit test) and optimized a bit, so it currently comes to 18 bytes. The serializer implementation could be simplified a bit if ~30 bytes is still an acceptable per-object overhead. (Truncating the `expires_at` timestamp to full-second integers saves 4 bytes per object, or I could use floating-points if the extra 4 bytes of precision is worth the tradeoff here.)
- I've added a subclass `::Dalli::E < Entry` to reduce the serialization overhead of the fully-qualified class name which is appended to the marshalled object (8 bytes for `Dalli::E`, compared to 39 for `ActiveSupport::Cache::DalliStore::Entry`).
- Using subclasses for the serialization implementation should also help with backwards-compatibility for cache stores persisting through version upgrades - e.g., if breaking changes ever need to be made to the Entry object's serialized format in the future, a differently-named subclass could be created. I don't know how important forwards- or backwards-compatibility is to Dalli's cache-store implementation (it might be a greater issue for Rails core, where Entry-object compatibility has been discussed extensively), but I thought I'd at least raise the issue and offer one potential solution in this PR.

lib/active_support/cache/dalli_store.rb
```diff
@@ -141,7 +135,17 @@ def write(name, value, options=nil)
         instrument(:write, name, options) do |payload|
           with do |connection|
             options = options.merge(:connection => connection)
-            write_entry(name, value, options)
+            # Support :race_condition_ttl by appending a secondary expiration to the value.
+            if options[:race_condition_ttl] &&
+              (ttl = options[:expires_in] || @options[:expires_in])
+              options[:expires_in] = ttl + options[:race_condition_ttl] * 2
```
petergoldstein: @wjordan Why are we adding twice the race_condition_ttl value to the expires_in here?

lib/active_support/cache/dalli_store.rb
```diff
@@ -141,7 +135,17 @@ def write(name, value, options=nil)
...
+              entry = Entry.new
```
petergoldstein: @wjordan This method now feels a lot longer and more complex.  I'd like to see it broken up a bit.  For example, the three lines generating the entry could easily be pulled out into a class method on the Entry class.  Similarly, the logic for the expires_in and ttl is difficult to follow as it's wrapped in the if condition.

djudd:
I'd love to be able to use this feature. Anything I can help with to move this forward?

wjordan:
My understanding based on #573 is that this class will soon be deprecated in favor of `ActiveSupport::Cache::MemCacheStore` (which already supports the feature implemented in this PR), so I didn't think it was worth moving this PR any further forward at this point.

@djudd if you have a use case for this PR that the existing `ActiveSupport::Cache::MemCacheStore` doesn't satisfy, I'd love to hear it.

djudd:
I don't, just didn't know that was happening. I'll look into switching instead. Thanks!

jarthod:
Hi, checking in on this I see it hasn't been merged but on the other hand the Readme doesn't say anything about `DalliStore` being deprecated in favor of `ActiveSupport::Cache::MemCacheStore` yet so what is considered the preferred way now? if we need `race_condition_ttl` we don't have much choice I guess?

PR #568, opened 2015-11-13
stop shipping files that are not needed to run/understand the code

grosser:
@mperham @petergoldstein

makes the gem lighter for those committing vendor/cache and install faster
leaving docs in case someone opens the gem up

### DIFF
--- a/dalli.gemspec
+++ b/dalli.gemspec
@@ -1,29 +1,23 @@
 require './lib/dalli/version'
 
 Gem::Specification.new do |s|
-  s.name = %q{dalli}
+  s.name = "dalli"
   s.version = Dalli::VERSION
   s.license = "MIT"
 
   s.authors = ["Mike Perham"]
-  s.description = %q{High performance memcached client for Ruby}
-  s.email = %q{mperham@gmail.com}
+  s.description = s.summary = "High performance memcached client for Ruby"
+  s.email = "mperham@gmail.com"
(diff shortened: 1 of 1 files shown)


dalli.gemspec
```diff
@@ -1,27 +1,21 @@
...
-  s.description = %q{High performance memcached client for Ruby}
-  s.email = %q{mperham@gmail.com}
+  s.description = s.summary = "High performance memcached client for Ruby"
+  s.email = "mperham@gmail.com"
   s.files = Dir.glob("lib/**/*") + [
      "LICENSE",
      "README.md",
      "History.md",
-     "Rakefile",
      "Gemfile",
```
petergoldstein: It looks like there's a trailing comma here that should be removed.

grosser: done!

petergoldstein:
@grosser Can you please rebase this PR?

Also, it looks like you've excluded the dalli.gemspec file from the included files.  That seems incorrect - can you clarify why you feel that's the correct change?

grosser:
a packed gem is a .gem file, which includes the gemspec and 'all other files', so adding the gemspec to the 'all other files' means it's included twice. I'm shipping lots of gems without it and it has never been an issue, even rails does not do it https://github.com/rails/rails/blob/master/rails.gemspec#L19

petergoldstein:
Fair enough, that makes sense.  Thanks for clarifying.

PR #569, opened 2015-11-13
use bundler and stop rescuing LoadError

grosser:
@mperham @petergoldstein


grosser:
@mperham @petergoldstein good to go ?

petergoldstein:
Thanks!

PR #570, opened 2015-11-13
test with and without local cache

grosser:
@mperham @petergoldstein

a preview of what I have in mind to tighten local cache testing, looks good ? -> I'll convert more tests and kick out the standalone LocaleCache tests

decided against using something that is inside the testcase so that you can dig into a specific issue with `ruby xxx_test.rb -n 'bla bla with LocaleCache'`

### DIFF
--- a/test/test_active_support.rb
+++ b/test/test_active_support.rb
@@ -9,6 +9,23 @@ class MockUser
 end
 
 describe 'ActiveSupport::Cache::DalliStore' do
+  # with and without local cache
+  def self.it_with_and_without_local_cache(message, &block)
+    it "#{message} with LocalCache" do
+      with_cache do
+        @dalli.with_local_cache do
+          instance_eval(&block)
+        end
+      end
+    end
+
+    it "#{message} without LocalCache" do
+      with_cache do
+        instance_eval(&block)
+      end
+    end
+  end
+
   describe 'active_support caching' do
     it 'has accessible options' do
       with_cache :expires_in => 5.minutes, :frob => 'baz' do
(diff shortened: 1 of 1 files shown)


test/test_active_support.rb
```diff
@@ -9,6 +9,23 @@ def cache_key
...
+    it "#{message} with LocaleCache" do
+      with_cache do
+        @dalli.with_local_cache do
+          instance_eval(&block)
+        end
+      end
+    end
+
+    it "#{message} without LocaleCache" do
+      with_cache do
```
petergoldstein: Looks like there's a typo in these two messages - 'LocaleCache'.  Plus I'm not really nuts about the method name - very obscure.  Can you come up with something clearer, although potentially more verbose?

grosser: `it_with_and_without_locale_cache` ?

petergoldstein: `it_with_and_without_local_cache` (same typo) would certainly be clearer.

grosser: dammit ... I do this typo way too often ... working to much in i18n ;D

grosser:
updated all that made sense and did not fail

grosser:
best viewed [without whitespace](https://github.com/mperham/dalli/pull/570/files?w=1)

grosser:
@petergoldstein

grosser:
@petergoldstein does this look good ?

grosser:
@mperham @petergoldstein can you please review this ?

petergoldstein:
@grosser Conceptually this seems fine.  If you'd like to work it into a full PR I'd be happy to take a look.  Thanks!

grosser:
I think it's done now, all tests converted that are make sense to run with and without.

petergoldstein:
Thanks!

PR #528, opened 2015-03-24
Log retryable errors with 'warn'

phrinx:

PR #578, opened 2015-12-07
Shift logging to warn level.

petergoldstein:


petergoldstein:
Implements the same change as PR #528

Issue #526, opened 2015-03-17
Remove custom binary search in Ring


mperham:
Ruby 2.0 added Array#bsearch.  Dalli 3.0 could remove 1.9.x support and start using that method instead.

PR #589, opened 2015-12-16
Refactor Ring

nateberkopec:
Unless there is some plugin or other infrastructure that somehow extends Dalli::Server::Ring, we can greatly simplify the included code.

In any case, I don't see why re-implementing `bsearch` is necessary, maybe I missed something or am missing context. This version performs equally to the previous.

The continuum is now sorted in *descending* rather than *ascending* value order - not sure if anyone is depending on that.

### DIFF
--- a/lib/dalli/ring.rb
+++ b/lib/dalli/ring.rb
@@ -12,15 +12,15 @@ module Dalli
       @continuum = nil
       if servers.size > 1
         total_weight = servers.inject(0) { |memo, srv| memo + srv.weight }
-        continuum = []
+        @continuum = []
         servers.each do |server|
           entry_count_for(server, servers.size, total_weight).times do |idx|
             hash = Digest::SHA1.hexdigest("#{server.name}:#{idx}")
             value = Integer("0x#{hash[0..7]}")
-            continuum << Dalli::Ring::Entry.new(value, server)
+            @continuum << { value: value, server: server }
           end
         end
-        @continuum = continuum.sort { |a, b| a.value <=> b.value }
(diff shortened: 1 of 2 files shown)


lib/dalli/ring.rb
```diff
@@ -31,8 +31,8 @@ def server_for_key(key)
       if @continuum
         hkey = hash_for(key)
         20.times do |try|
-          entryidx = binary_search(@continuum, hkey)
-          server = @continuum[entryidx].server
+          entry = @continuum.bsearch { |h| h[:value] <= hkey } || @continuum.last
```
nateberkopec: This `|| @continuum.last` hack is weird - try removing it and seeing which tests fail. I think I messed up my implementation here somehow.

petergoldstein:
@nateberkopec Thanks for the contributions this morning.  Appreciate it.

Regarding this one in particular, there's an implementation of bsearch (rather than using the method on Ruby) because of a Ruby 1.9.3 compatibility requirement for the Sidekiq 2.7.x set of releases.  The tracked issue is #526 .

I'm expecting Dalli::Ring to get a lot of rewriting during the run up to 3.0, to support pluggable algorithms, etc.  And eliminating the custom bsearch will be part of that, as support for 1.9.3 is dropped.

Given the above, I'm not sure it makes sense to merge in this change now, outside of the work done for the larger context.

nateberkopec:
> because of a Ruby 1.9.3 compatibility requirement for the Sidekiq 2.7.x set of releases

Aha, knew it was too obvious a change!

PR #594, opened 2016-01-25
Allow Dalli::Client#incr and #decr to take negative numbers

mwpastore:
A simple patch to reroute `incr` to `decr` and `decr` to `incr` when negative increments are encountered.

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -161,7 +161,6 @@ module Dalli
 
     ##
     # Incr adds the given amount to the counter on the memcached server.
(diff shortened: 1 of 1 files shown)


petergoldstein:
I'm inclined not to merge this.  We'd be changing the contracts of public methods, for little to no real benefit.

As a general rule a caller should know whether the value being passed to increment or decrement is positive or negative.  And if the caller really wants this behavior then they can wrap incr and decr in their own methods.

mwpastore:
Fair enough. I ended up doing just that where I needed it. Thanks Peter.

PR #597, opened 2016-02-01
Add #cas! method that yields whether or not the key exists

mwpastore:
Looking for a variant of the #cas method that yields to the block whether or not the key already exists. This is to avoid having to do things like:

```ruby
result = dc.cas(key) { |oldvalue| newvalue }
result = dc.add(key, newvalue) if result.nil?
```

Thanks in advance.

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -107,13 +107,19 @@ module Dalli
     # - nil if the key did not exist.
     # - false if the value was changed by someone else.
     # - true if the value was successfully updated.
-    def cas(key, ttl=nil, options=nil)
-      (value, cas) = perform(:cas, key)
-      value = (!value || value == 'Not found') ? nil : value
-      if value
-        newvalue = yield(value)
-        perform(:set, key, newvalue, ttl_or_default(ttl), cas, options)
-      end
+    def cas(key, ttl=nil, options=nil, &block)
+      cas_core(key, false, ttl, options, &block)
+    end
+
+    ##
+    # like #cas, but will yield to the block whether or not the value
+    # already exists.
+    #
+    # Returns:
+    # - false if the value was changed by someone else.
+    # - true if the value was successfully updated.
+    def cas!(key, ttl=nil, options=nil, &block)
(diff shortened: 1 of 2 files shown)


petergoldstein:
So I haven't given this PR any kind of detailed review yet, but at a minimum it appears that the code is being added in the wrong place.  CAS related code should be in lib/dalli/cas/client.rb

mwpastore:
Hi Peter, Dalli::Client#cas is located in lib/dalli/client.rb so I added #cas! there for consistency. Would you like me to move both methods to lib/dalli/cas/client.rb if that's the correct location for them?

petergoldstein:
Actually I think you're right - my apologies.  I'll take a look at the PR sometime this week.

mwpastore:
Sounds good. Thank you!

petergoldstein:
Two quick pieces of feedback:

1. This PR needs specs.  You've added functionality, but no corresponding tests.  See the 'support the cas operation' section in `test/test_dalli.rb` for the `cas` method tests.  You should add similar tests for`cas!`, and ensure that you specifically test the difference in behavior between the two methods.

2. This PR introduces a lot of code duplication which should be pulled out into a common (private) method.  Something like:

```
def cas(key, ttl=nil, options=nil)
  cas_core(key, false, ttl, options)
end

def cas!(key, ttl=nil, options=nil)
  cas_core(key, true, ttl, options)
end

private

def cas_core(key, always_set, ttl=nil, options=nil)
  (value, cas) = perform(:cas, key)
  value = (!value || value == 'Not found') ? nil : value
  return if value.nil? && !always_set
  newvalue = yield(value)
  perform(:set, key, newvalue, ttl_or_default(ttl), cas, options)
end

```

Feel free to alter the above as you wish, but I think that represents the general idea.

mwpastore:
Ah, excellent feedback. Thanks Peter. I made the changes as suggested and added a test.

petergoldstein:
Thanks!

mwpastore:
Thanks Peter!

Issue #562, opened 2015-10-06
Rack::Session::Dalli with a ConnectionPool?


mwpastore:
Sorry for the vague non-issue, but I'm trying to determine how I can go about coercing/tweaking `rack/session/dalli.rb` to take a `ConnectionPool` object. I've tried passing it in as `:cache` but `Rack::Session::Dalli` expects (as I'm sure you're aware) a `Dalli::Client` that responds to the `#alive!`, `#set()`, `#delete()`, etc. methods. Should I simply use a `ConnectionPool::Wrapper`? Or is there a better way to go about this?

mwpastore:
Also, is `Rack::Session::Dalli` thread-safe?

petergoldstein:
When running in an environment where `env['rack.multithread']` is set to true, Rack::Session::Dalli will be thread-safe.  And unless explicitly disabled with the `threadsafe: false`  option a `Dalli::Client` will be thread-safe.

mwpastore:
Excellent. Thanks @petergoldstein. Any thoughts on using Rack::Session::Dalli with a ConnectionPool?

axelson:
Shouldn't this be closed by #598?

mwpastore:
@axelson Yup. Thank you!

PR #598, opened 2016-02-12
Modify Rack::Session::Dalli to be compatible with ConnectionPool

mwpastore:
I tried do this with as small of a footprint as possible. Let me know if you'd like me to take a different approach, e.g. with argument checking, or eliminate the `with_lock` method entirely if it's a ConnectionPool. Thank you!

n.b. This addresses #562!

### DIFF
--- a/lib/rack/session/dalli.rb
+++ b/lib/rack/session/dalli.rb
@@ -5,75 +5,164 @@ require 'dalli'
 module Rack
   module Session
     class Dalli < defined?(Abstract::Persisted) ? Abstract::Persisted : Abstract::ID
-      attr_reader :pool, :mutex
+      attr_reader :pool
 
-      DEFAULT_OPTIONS = Abstract::ID::DEFAULT_OPTIONS.merge \
+      DEFAULT_DALLI_OPTIONS = {
         :namespace => 'rack:session',
         :memcache_server => 'localhost:11211'
+      }
 
+      # Brings in a new Rack::Session::Dalli middleware with the given
+      # `:memcache_server`. The server is either a hostname, or a
+      # host-with-port string in the form of "host_name:port", or an array of
+      # such strings. For example:
+      #
+      #   use Rack::Session::Dalli,
+      #     :memcache_server => "mc.example.com:1234"
+      #
+      # If no `:memcache_server` option is specified, Rack::Session::Dalli will
+      # connect to localhost, port 11211 (the default memcached port). If
+      # `:memcache_server` is set to nil, Dalli::Client will look for
+      # ENV['MEMCACHE_SERVERS'] and use that value if it is available, or fall
+      # back to the same default behavior described above.
+      #
+      # Rack::Session::Dalli is intended to be a drop-in replacement for
+      # Rack::Session::Memcache. It accepts additional options that control the
+      # behavior of Rack::Session, Dalli::Client, and an optional
+      # ConnectionPool. First and foremost, if you wish to instantiate your own
+      # Dalli::Client (or ConnectionPool) and use that instead of letting
+      # Rack::Session::Dalli instantiate it on your behalf, simply pass it in
+      # as the `:cache` option. Please note that you will be responsible for
+      # setting the namespace and any other options on Dalli::Client.
+      #
+      # Secondly, if you're not using the `:cache` option, Rack::Session::Dalli
+      # accepts the same options as Dalli::Client, so it's worth reviewing its
+      # documentation. Perhaps most importantly, if you don't specify a
+      # `:namespace` option, Rack::Session::Dalli will default to using
+      # "rack:session".
+      #
+      # Whether you are using the `:cache` option or not, it is not recommend
+      # to set `:expires_in`. Instead, use `:expire_after`, which will control
+      # both the expiration of the client cookie as well as the expiration of
+      # the corresponding entry in memcached.
+      #
+      # Rack::Session::Dalli also accepts a host of options that control how
+      # the sessions and session cookies are managed, including the
+      # aforementioned `:expire_after` option. Please see the documentation for
+      # Rack::Session::Abstract::Persisted for a detailed explanation of these
+      # options and their default values.
+      #
+      # Finally, if your web application is multithreaded, the
+      # Rack::Session::Dalli middleware can become a source of contention. You
+      # can use a connection pool of Dalli clients by passing in the
+      # `:pool_size` and/or `:pool_timeout` options. For example:
+      #
+      #   use Rack::Session::Dalli,
+      #     :memcache_server => "mc.example.com:1234",
+      #     :pool_size => 10
+      #
+      # You must include the `connection_pool` gem in your project if you wish
+      # to use pool support. Please see the documentation for ConnectionPool
+      # for more information about it and its default options (which would only
+      # be applicable if you supplied one of the two options, but not both).
+      #
       def initialize(app, options={})
+        # Parent uses DEFAULT_OPTIONS to build @default_options for Rack::Session
         super
-        @mutex = Mutex.new
-        mserv = @default_options[:memcache_server]
-        mopts = @default_options.reject{|k,v| !DEFAULT_OPTIONS.include? k }
-        @pool = options[:cache] || ::Dalli::Client.new(mserv, mopts)
-        @pool.alive!
+
+        # Determine the default TTL for newly-created sessions
+        @default_ttl = ttl @default_options[:expire_after]
+
+        # Normalize and validate passed options
+        cache, mserv, mopts, popts = extract_dalli_options options
+
+        @pool =
+          if cache # caller passed a Dalli::Client or ConnectionPool instance
+            cache
+          elsif popts # caller passed ConnectionPool options
+            ConnectionPool.new(popts) { ::Dalli::Client.new(mserv, mopts) }
+          else
+            ::Dalli::Client.new(mserv, mopts)
+          end
+
+        @pool.alive! if @pool.respond_to? :alive!
       end
 
-      if defined?(Abstract::Persisted)
-        def find_session(req, sid)
-          get_session req.env, sid
-        end
-
-        def write_session(req, sid, session, options)
-          set_session req.env, sid, session, options
-        end
-
-        def delete_session(req, sid, options)
-          destroy_session req.env, sid, options
-        end
-      end
-
-      def generate_sid
-        while true
-          sid = super
-          break sid unless @pool.get(sid)
-        end
-      end
-
-      def get_session(env, sid)
-        with_lock(env, [nil, {}]) do
-          unless sid and !sid.empty? and session = @pool.get(sid)
-            sid, session = generate_sid, {}
-            unless @pool.add(sid, session)
-              raise "Session collision on '#{sid.inspect}'"
+      def get_session(_, sid)
+        with_block([nil, {}]) do |dc|
+          unless sid and !sid.empty? and session = dc.get(sid)
+            old_sid, sid, session = sid, generate_sid_with(dc), {}
+            unless dc.add(sid, session, @default_ttl)
+              sid = old_sid
+              redo # generate a new sid and try again
             end
           end
           [sid, session]
         end
       end
 
-      def set_session(env, session_id, new_session, options)
+      def set_session(_, session_id, new_session, options)
         return false unless session_id
-        expiry = options[:expire_after]
-        expiry = expiry.nil? ? 0 : expiry + 1
 
-        with_lock(env, false) do
-          @pool.set session_id, new_session, expiry
+        with_block(false) do |dc|
+          dc.set(session_id, new_session, ttl(options[:expire_after]))
           session_id
         end
       end
 
-      def destroy_session(env, session_id, options)
-        with_lock(env) do
-          @pool.delete(session_id)
-          generate_sid unless options[:drop]
+      def destroy_session(_, session_id, options)
+        with_block do |dc|
+          dc.delete(session_id)
+          generate_sid_with(dc) unless options[:drop]
         end
       end
 
-      def with_lock(env, default=nil)
-        @mutex.lock if env['rack.multithread']
-        yield
+      if defined?(Abstract::Persisted)
+        # We don't need to destructure the request, because we're not using
+        # request.env in the methods above, so simply alias the method names.
+        alias_method :find_session, :get_session
+        alias_method :write_session, :set_session
+        alias_method :delete_session, :destroy_session
+      end
+
+      private
+
+      def extract_dalli_options(options)
+        return [options[:cache]] if options[:cache]
+
+        # Filter out Rack::Session-specific options and apply our defaults
+        mopts = DEFAULT_DALLI_OPTIONS.merge \
+          options.reject {|k, _| DEFAULT_OPTIONS.key? k }
+        mserv = mopts.delete :memcache_server
+
+        if mopts[:pool_size] || mopts[:pool_timeout]
+          popts = {}
+          popts[:size] = mopts.delete :pool_size if mopts[:pool_size]
+          popts[:timeout] = mopts.delete :pool_timeout if mopts[:pool_timeout]
+
+          # For a connection pool, locking is handled at the pool level
+          mopts[:threadsafe] = false unless mopts.key? :threadsafe
+        end
+
+        [nil, mserv, mopts, popts]
+      end
+
+      # Capture generate_sid's super so we can call it from generate_sid_with
+      alias_method :generate_sid_super, :generate_sid
+
+      def generate_sid
+        with_block {|dc| generate_sid_with(dc) }
+      end
+
+      def generate_sid_with(dc)
+        while true
+          sid = generate_sid_super
+          break sid unless dc.get(sid)
+        end
+      end
+
+      def with_block(default=nil, &block)
+        @pool.with(&block)
       rescue ::Dalli::DalliError, Errno::ECONNREFUSED
         raise if $!.message =~ /undefined class/
         if $VERBOSE
@@ -81,10 +170,11 @@ module Rack
           warn $!.inspect
         end
         default
-      ensure
-        @mutex.unlock if @mutex.locked?
       end
 
+      def ttl(expire_after)
+        expire_after.nil? ? 0 : expire_after + 1
+      end
     end
   end
 end
--- a/lib/rack/session/dalli.rb
+++ b/lib/rack/session/dalli.rb
@@ -5,7 +5,7 @@ require 'dalli'
 module Rack
   module Session
     class Dalli < defined?(Abstract::Persisted) ? Abstract::Persisted : Abstract::ID
-      attr_reader :pool
+      attr_reader :pool, :mutex
 
       DEFAULT_DALLI_OPTIONS = {
         :namespace => 'rack:session',
@@ -85,11 +85,15 @@ module Rack
             ::Dalli::Client.new(mserv, mopts)
           end
 
-        @pool.alive! if @pool.respond_to? :alive!
+        if @pool.respond_to?(:alive!) # is a Dalli::Client
+          @mutex = Mutex.new
+
+          @pool.alive!
+        end
       end
 
-      def get_session(_, sid)
-        with_block([nil, {}]) do |dc|
+      def get_session(env, sid)
+        with_block(env, [nil, {}]) do |dc|
           unless sid and !sid.empty? and session = dc.get(sid)
             old_sid, sid, session = sid, generate_sid_with(dc), {}
             unless dc.add(sid, session, @default_ttl)
@@ -101,28 +105,34 @@ module Rack
         end
       end
 
-      def set_session(_, session_id, new_session, options)
+      def set_session(env, session_id, new_session, options)
         return false unless session_id
 
-        with_block(false) do |dc|
+        with_block(env, false) do |dc|
           dc.set(session_id, new_session, ttl(options[:expire_after]))
           session_id
         end
       end
 
-      def destroy_session(_, session_id, options)
-        with_block do |dc|
+      def destroy_session(env, session_id, options)
+        with_block(env) do |dc|
           dc.delete(session_id)
           generate_sid_with(dc) unless options[:drop]
         end
       end
 
       if defined?(Abstract::Persisted)
-        # We don't need to destructure the request, because we're not using
-        # request.env in the methods above, so simply alias the method names.
-        alias_method :find_session, :get_session
-        alias_method :write_session, :set_session
-        alias_method :delete_session, :destroy_session
+        def find_session(req, sid)
+          get_session req.env, sid
+        end
+
+        def write_session(req, sid, session, options)
+          set_session req.env, sid, session, options
+        end
+
+        def delete_session(req, sid, options)
+          destroy_session req.env, sid, options
+        end
       end
 
       private
@@ -151,7 +161,9 @@ module Rack
       alias_method :generate_sid_super, :generate_sid
 
       def generate_sid
-        with_block {|dc| generate_sid_with(dc) }
+        # no way to check env['rack.multithread'] here so fall back on
+        # Dalli::Client or ConnectionPool's internal mutex cf. our own
+        @pool.with {|dc| generate_sid_with(dc) }
       end
 
       def generate_sid_with(dc)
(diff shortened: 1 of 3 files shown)


mwpastore:
I don't know why that one test failed, but it appears to be a transient error:

```
failed to listen on TCP port 34547: Address already in use
failed to listen on TCP port 34541: Address already in use
```

petergoldstein:
So I think is a reasonable first pass, but I think the initialization phase doesn't cover the cases we probably want it to cover and isn't entirely consistent with how connection pool support is implemented elsewhere.  Specifically here - https://github.com/petergoldstein/dalli/commit/c51ce5d826b3556cb9533ff3246277613d10dc2c

My take is that you probably want to handle four scenarios, in the following order:

1. Caller passes in pool_options which, if present, cause a ConnectionPool to be instantiated.
2. Caller passes in a ConnectionPool instance
3. Caller passes in a Dalli::Client instance
4. Caller passes in nothing, and the initialize method instantiates a Dalli::Client for internal use

The current form of the PR doesn't allow option 1, which seems incongruous considering that Dalli explicitly supports that in the Rails cache implementation.

I'd suggest breaking the `initialize` method into pieces to handle each of the 4 options above, and set up the @pool and @mutex variables as needed.

What do you think?

mwpastore:
Let me start by saying that I think these are all good ideas and I'm certainly willing to put in the effort. 

However, it would involve quite a significant rewrite of Rack::Session::Dalli's constructors. Currently, the only attributes of the Dalli::Client object that can be customized when instantiating Rack::Session::Dalli are `:memcache_server` and `:namespace`. It doesn't support any of the other options, like `:failover` or `:serializer`; it looks like it was written to be explicitly, 1:1 compatible with Rack::Session::Memcache. Currently, the only way to customize the Dalli::Client object are to instantiate it yourself ahead of time and pass it in as the (undocumented? but widely referenced online) `:cache` option. To maintain a stable interface, the best option is to simply allow the user to pass in a ConnectionPool object instead of a Dalli::Client object, which is what I've done here. Rack::Session::Dalli has a completely different ancestry and use case than ActiveSupport::Cache::DalliStore so I'm not sure it makes sense to look there for a model implementation, but I certainly understand the desire to be consistent.

All that being said, if you're comfortable with me embarking on this overhaul, possibly changing the inheritance model, and possibly breaking compatibility with Rack::Session::Memcache, I will table this PR for now and come back to it in a few weeks when I have more time.

mwpastore:
Argh, okay, let me walk that back a little. I didn't understand how the Abstract::ID parent class was manipulating `DEFAULT_OPTIONS` and `@default_options`. You definitely _can_ pass in e.g. `:compress` and it will get passed to Dalli::Client.new. Let me play with this a bit... maybe it won't be as bad as I think.

petergoldstein:
Ok, I thought your initial comments were a little strong.  I'm unclear why you can't use default_options to handle the cases I describe, in a way that is completely backwards compatible, but supports the additional cases with other option flags.  That's essentially what's being done with Dalli::Client now.

mwpastore:
They were a little strong, and I apologize. I'm a dope. Thanks Peter.

mwpastore:
Here's a question that maybe you can answer: Why does this thing even have a mutex in the first place? Doesn't Dalli::Client do its own thread-safety? Is it possibly a holdover from before it was thread-safe (and therefore we can remove it)?

petergoldstein:
That's actually an excellent question, and I don't really know the answer.  From `blame` we can see that it was added in PR #264 , which post-dates the introduction of Dalli thread safety.  From the comments in the PR, it appears to be driven by concerns over the removal of Rack::Lock in Rails 4.  And the code seems to be almost a straight port of https://github.com/rack/rack/blob/master/lib/rack/session/memcache.rb

But as you note, Dalli::Client is threadsafe by default, so my suspicion is that the mutex (and the corresponding complexity around it) is probably unnecessary.  And the same would be true with a ConnectionPool.

I don't have time to look into this right now, but I suspect there's a little cargo cultism going on here, and we could eliminate the locking logic entirely.

mwpastore:
Something is still not quite right with this options parsing logic. I think there might have been a hint of truth to what I was trying to say in my mini-rant earlier. If you pass in e.g. `:compress => true` as an option to Rack::Session::Dalli, it will be present in `@default_options`, but [this line](https://github.com/petergoldstein/dalli/blob/v2.7.5/lib/rack/session/dalli.rb#L17) will exclude it from `mopts` (because it's not present in `DEFAULT_OPTIONS`), and it won't be passed to Dalli::Client.new. The filter should really be by the options that Dalli::Client.new accepts, plus the pool options we've just introduced. I'm working on it...

mwpastore:
Okay, I'm feeling pretty confident about this. Take a look. I've added a documentation block as well, and I made some assertions in there that may not be true, so a sanity check would be appreciated. Thanks Peter!

petergoldstein:
@mwpastore Appreciate the effort.  I'm unlikely to get a chance to review today, but I'll try to get to it this week.

Also, there have been some changes since this was pushed - Rails 5 compatibility issues.  Can you rebase this branch when you have a minute?

Thanks

mwpastore:
Done!

PetrKaleta:
This is exactly what I was looking for. I hope it will be merged soon

lib/rack/session/dalli.rb
```diff
@@ -4,75 +4,143 @@
...
-          get_session req.env, sid
-        end
+        # determine the default TTL for newly-created sessions
+        @default_ttl = ttl @default_options[:expire_after]
 
-        def write_session(req, sid, session, options)
-          set_session req.env, sid, session, options
-        end
+        mopts = DEFAULT_DALLI_OPTIONS.merge \
+          options.reject {|k, _| DEFAULT_OPTIONS.key? k }
```
petergoldstein: So there appears to be a `!` missing when compared to the original code.  @mwpastore this looks like an error to me.  Can you please comment?

Frankly I'm not sure why the original code had the `reject` and `!` structure, instead of using a more semantically clear `select`.  If there's no real reason for that, I'd suggest the latter.

lib/rack/session/dalli.rb
```diff
@@ -4,75 +4,143 @@
...
-      def generate_sid
-        while true
-          sid = super
-          break sid unless @pool.get(sid)
-        end
+        @pool =
+          if options[:cache] and options[:cache].respond_to? :with
+            options[:cache]
+          elsif !pool_options.empty?
+            mopts[:threadsafe] = false unless mopts.key? :threadsafe
```
petergoldstein: Why are we defaulting to `threadsafe` false here?

petergoldstein:
@mwpastore Sorry it's taken me a few weeks to get back to this.

On a high level, I think the approach looks ok.  I left a few line comments that need to be addressed before we can discuss merging.

As per our earlier conversation, I suspect that the mutexes were there from a cargo cult change, and not because they were actually required.  So I'm fine with removing them, provided you do a bit of concurrency sanity checking beforehand.

It looks like you covered the cases I mentioned in my earlier comment, albeit in a slightly different order (2/3, 1, 4).  I think that ordering is probably fine.

One key item that's missing is any sort of testing.  You're going to need to add some tests to cover this new functionality, and to ensure options are being handled as expected.

Finally, I'd still recommend breaking up the initialize method or, at the very least, adding a couple of syntactic sugar methods to clarify exactly what is being tested and done in the `if` clause used to create the `@pool` object

lib/rack/session/dalli.rb
mwpastore: I'm replicating the behavior added to ActiveSupport::Cache::DalliStore in the commit you linked me to earlier: https://github.com/petergoldstein/dalli/commit/c51ce5d826b3556cb9533ff3246277613d10dc2c

lib/rack/session/dalli.rb
mwpastore: I had to fundamentally altered the options parsing logic of Rack::Session::Dalli. The original version leverages Rack::Session::Abstract::Persisted's setup to handle the options for both Rack::Session::Abstract::Persisted and Dalli::Client, using `DEFAULT_OPTIONS` as a whitelist for both. Any options not added to the whitelist, such as `:failover` or `:serializer`, will not be passed to Dalli::Client. This severely limits the ability of the user to customize the Dalli::Client instance being used by Rack::Session::Dalli; only `:namespace` and the Memcached server can be specified! It also prevents us from passing `:pool_size` and `:pool_timeout` options through to the ConnectionPool setup.

One solution would have been to add the Dalli::Client and ConnectionPool options that we're "okay" with to `DEFAULT_OPTIONS`. I started doing this and after reviewing the 19 additional options available to Dalli::Client and adding them to the list, plus the two for ConnectionPool, it occurred to me that this is kind of silly. In my mind, any options the user passes to Rack::Session::Dalli at `use` time should pass to Rack::Session::Abstract::Persisted first, and any leftover options should pass to Dalli::Client, with each class handling its own setup. We can apply any defaults specific to Rack::Session::Dalli in our logic (such as the default namespace), and peel off the Memcached server name(s) and any ConnectionPool options for "special" handling.

That's what the new options parsing does:

1. Call `super`. Rack::Session::Abstract::Persisted will use `DEFAULT_OPTIONS` to build `@default_options`. These are options that primarily control how the session cookie is generated.
1. Remove any elements from the options hash that are/were handled by Rack::Session::Abstract::Persisted, leaving only the Memcached server and options for Dalli::Client and ConnectionPool.
1. Apply the defaults specified in `DEFAULT_DALLI_OPTIONS` to the remaining options.
1. Remove the Memcached server name(s) and ConnectionPool options (as these shouldn't be passed to Dalli::Client in the options hash). 

So that's why we're rejecting instead of selecting options based on `DEFAULT_OPTIONS`. It's no longer a whitelist; it's now a way of differentiating between options intended for Rack::Session::Abstract::Persisted and options intended for Dalli::Client and ConnectionPool.

mwpastore:
@petergoldstein No worries. Thank you for taking the time today to review it. I will work on adding some additional tests for the new functionality and breaking up `initialize`.

lib/rack/session/dalli.rb
petergoldstein: So that's a really clear explanation.  And without it, I'd have no idea what's going on in this line (or more likely, would misinterpret it).  And the odds of someone working on this code going and digging up that line comment are near zero.

In my mind that makes this a very good candidate for a method extraction with a condensed version of the above as a comment.  Some sort of `extract_dalli_options` method that returns `mopts`, `mserv`, `pool_options`, and conditionally `options[:cache]` (when it responds to `with`) would be a lot more comprehensible 6-12 months from now.  That would also allow a substantially simplification of the `if` clause that builds the `@pool`

lib/rack/session/dalli.rb
petergoldstein: Ah, got it.  This only applies in the pool case, when the lock is held at the pool level.

Again, if moved to the proposed `extract_dalli_options` method described above, with an appropriate line comment (e.g. 'For a connection pool, locking is handled at the pool level') this would be a lot clearer.

mwpastore: The original version had zero line comments, so I was just preserving the code style. :stuck_out_tongue: I'll clean it up. Thanks Peter.

mwpastore:
@petergoldstein Alrighty, take another look when you can and let me know if you'd like to see any more changes or tests!

mwpastore:
Hi Peter, just a heads up that I found and fixed a new bug yesterday. I had changed the calling signature of `generate_sid` to avoid nesting `@pool.with` blocks, not realizing that `generate_sid` is called directly from Rack::Session::Abstract::Persisted [here](https://github.com/rack/rack/blob/1.6.4/lib/rack/session/abstract/id.rb#L332) if `:renew` or `:drop` are specified. I think I found a solution that lets us have the best of both worlds; the relevant line changes are [here](https://github.com/mwpastore/dalli/blob/rack-session/lib/rack/session/dalli.rb#L149-L154). I also added a test to prevent this from tripping up someone else in the future. Let me know if you have any questions.

mwpastore:
## Benchmark Results

Here are some preliminary benchmark results comparing the performance of Rack::Session::Dalli under heavy thread concurrency before and after this patch, and with and without connection pooling. Testing was performed using wrk against a simple Sinatra app running on Puma 3.1/JRuby 9.0.5.0/OpenJDK 7u95. The benchmark is far from comprehensive but it does a get and a set on every request. My code is [here](https://github.com/mwpastore/dalli-bench).

### Without Connection Pooling

#### Before (Dalli 2.7.6 release)

```
Running 1m test @ http://localhost:9292
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   354.10ms  101.44ms 976.56ms   69.19%
    Req/Sec    91.84     29.26   323.00     70.14%
  65961 requests in 1.00m, 15.23MB read
Requests/sec:   1097.82
Transfer/sec:    259.60KB
```

#### After

```
Running 1m test @ http://localhost:9292
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   381.95ms  137.87ms 969.73ms   69.78%
    Req/Sec    84.80     27.88   200.00     72.44%
  60879 requests in 1.00m, 14.06MB read
Requests/sec:   1012.99
Transfer/sec:    239.58KB
```

#### Delta

Avg Latency: 1.08x
Requests/sec: 0.92x

Unfortunately, it looks like this new version is about 8% slower than the current release (for this simple benchmark, at least). I'd like to write some different benchmarks and do some additional profiling but I think it's safe to assume that calling Dalli::Client#with with a block is just simply slower than calling the #get, #set, #add, and #delete methods directly. It probably would be even worse if we hadn't also removed the unnecessary mutex. 

### With Connection Pooling

```
Running 1m test @ http://localhost:9292
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    98.59ms   47.54ms 453.85ms   58.51%
    Req/Sec   336.16     48.45   707.00     74.34%
  241251 requests in 1.00m, 55.72MB read
Requests/sec:   4014.28
Transfer/sec:      0.93MB
```

#### Delta (cf. Dalli 2.7.6 release)

Avg Latency: 0.28x
Requests/sec: 3.96x

That's more like it! So the question becomes whether the the huge speedup with connection pooling justifies the slight slowdown without.

petergoldstein:
So I wouldn't assume this without testing - "it's safe to assume that calling Dalli::Client#with with a block is just simply slower than calling the #get, #set, #add, and #delete methods directly".  You may in fact be right, but it seems like this would be an easy thing to benchmark and confirm or refute.

That said, I'm inclined to pay the performance hit on the simple case (especially if we can shave a few more % off with some optimization) for the scalability/performance benefit of the connection pool case.

Also, can you give some idea of how performance scales with # of connections?  What happens if you run the same test again, but with 20 connections?  10 connections?  Are 400 connections even useful in a test with 12 threads?

mwpastore:
### with vs. without with

I wrote a simple benchmark to simulate a session flow for a single request with `:renew` set to true (which is what the main benchmark does) and a bad rack.session cookie on the request. Some of the ops will occur within the same call to with_block, others will take place in their own. Code is up on the same Git repository.

```
        without with   132.000  i/100ms
                with   152.000  i/100ms
        without with      1.672k (± 5.8%) i/s -     50.028k
                with      1.649k (± 6.4%) i/s -     49.248k
```

> I wouldn't assume this without testing

As you suspected, my assumptions did not bear out, so I will have to do some more profiling to figure out where the performance hit is coming from. #with only accounts for about 1.6% of the difference. 

### wrk revisited

> Are 400 connections even useful in a test with 12 threads?

I'm not sure. This is my first experience with wrk and the documentation is... scant. My understanding is that the threads mostly come into play when you are running Lua scripts to simulate more complex sequences of requests, and that simply juggling 400 open connections requires significantly less horsepower. wrk was never a huge consumer of CPU (compared to Ruby and Memcached) in my testing, even when there were ample cycles available for it to consume. 

I kept the threads at 12 and ran some tests to show the scaling. I couldn't do 10 because wrk requires at least as many connections as there are threads. I ran these on a differently-sized box than the above so I repeated the 400 connections test as well. Let me know what else you'd like to see. 

#### 12 connections

```
Running 1m test @ http://localhost:9292
  12 threads and 12 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    40.21ms    1.30ms  60.02ms   93.26%
    Req/Sec    24.88      5.00    30.00     51.12%
  17914 requests in 1.00m, 4.14MB read
Requests/sec:    298.28
Transfer/sec:     70.53KB
```

#### 20 connections

```
Running 1m test @ http://localhost:9292
  12 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    40.25ms    1.30ms  53.31ms   92.06%
    Req/Sec    24.85      5.00    30.00     51.43%
  17895 requests in 1.00m, 4.13MB read
Requests/sec:    297.96
Transfer/sec:     70.46KB
```

#### 40 connections

```
Running 1m test @ http://localhost:9292
  12 threads and 40 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    40.72ms    1.79ms  60.07ms   79.61%
    Req/Sec    73.68      9.22    90.00     69.15%
  53053 requests in 1.00m, 12.25MB read
Requests/sec:    883.57
Transfer/sec:    208.95KB
```

#### 80 connections

```
Running 1m test @ http://localhost:9292
  12 threads and 80 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    41.83ms    2.55ms  70.88ms   90.71%
    Req/Sec   143.90     13.12   181.00     62.00%
  103304 requests in 1.00m, 23.86MB read
Requests/sec:   1720.18
Transfer/sec:    406.79KB
```

#### 160 connections

```
Running 1m test @ http://localhost:9292
  12 threads and 160 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    49.19ms    8.88ms 208.80ms   72.72%
    Req/Sec   264.05     27.33   350.00     77.64%
  189601 requests in 1.00m, 43.79MB read
Requests/sec:   3154.89
Transfer/sec:    746.10KB
```

#### 320 connections

```
Running 1m test @ http://localhost:9292
  12 threads and 320 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    92.53ms   45.91ms 651.36ms   62.75%
    Req/Sec   282.63     44.76   470.00     72.11%
  202886 requests in 1.00m, 46.86MB read
Requests/sec:   3375.87
Transfer/sec:    798.40KB
```

#### 400 connections

```
Running 1m test @ http://localhost:9292
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   114.08ms   64.03ms 524.25ms   58.39%
    Req/Sec   291.81     53.37   717.00     73.91%
  209490 requests in 1.00m, 48.37MB read
Requests/sec:   3485.86
Transfer/sec:    824.24KB
```

#### 800 connections

```
Running 1m test @ http://localhost:9292
  12 threads and 800 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   111.83ms   63.05ms 492.30ms   61.15%
    Req/Sec   296.34    166.81   730.00     62.88%
  212895 requests in 1.00m, 49.18MB read
Requests/sec:   3542.40
Transfer/sec:    837.90KB
```

dalli.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'minitest', '>= 4.2.0'
   s.add_development_dependency 'mocha'
   s.add_development_dependency 'rails', '~> 4'
-  s.add_development_dependency 'rake'
+  s.add_development_dependency 'rake', '~> 10.5'
```
petergoldstein: What's the reason for locking the development dependency to '~> 10.5'?

mwpastore: Rake 11 appears to break JRuby on Travis.

petergoldstein: I think this is fixed in Rake 11.1.1

mwpastore: Ah, excellent. Would you like me to set it to `>= 11.1.1`, or should I revert the change entirely?

petergoldstein: Just revert this change.

mwpastore: Done.

petergoldstein:
> I will have to do some more profiling to figure out where the performance hit is coming from. #with only accounts for about 1.6% of the difference.

@mwpastore Did you ever do any of this profiling?

mwpastore:
I have not, is that something you need to see before merging this PR?

petergoldstein:
Yes.

In your original benchmarking you saw an 8% performance hit.  As I said in response:

> I'm inclined to pay the performance hit on the simple case (especially if we can shave a few more % off with some optimization) for the scalability/performance benefit of the connection pool case.

We can't shave off those few more % unless we know where the performance hit is coming from.

mwpastore:
Ah, I thought you were saying you were fine with the performance hit because of the overwhelming benefits of connection pooling. I'll get back into it.

mwpastore:
After profiling both versions, here's what jumps out:

base:

```ruby
  "ConditionVariable#wait": {
    "id": 1596,
    "total_calls": 801,
    "total_time": 5381.941433999996,
    "self_time": 5381.941433999996,
    "child_time": 0.0
  }
```

head:

```ruby
  "ConditionVariable#wait": {
    "id": 1596,
    "total_calls": 797,
    "total_time": 9928.871749000005,
    "self_time": 9928.871749000005,
    "child_time": 0.0
  }
```

In both versions, `ConditionVariable#wait` is only called by `Mutex#synchronize`. If you'll recall, this PR removes the mutex within Rack::Session::Dalli in favor of Dalli::Client's own thread safety mechanism, which appears to be less performant for reasons beyond my understanding. We can either add the mutex back in to use when `@pool` is not a ConnectionPool—possibly setting `:threadsafe` to false on the Dalli::Client in either case—or we can leave it as-is and eat the ~7% performance penalty.

petergoldstein:
@mwpastore Sorry - just saw your last comment.  Let me think about this a bit.  It's more than a little strange that we're seeing such a difference in synchronization performance.

mwpastore:
@petergoldstein It's bothering me too. I wrote another, even simpler benchmark that just does `#set` and a `#get` with and without a surrounding mutex and my findings bear out. It looks like the Monitor used by Dalli::Threadsafe isn't as "efficient" as the simple Mutex employed by the old version of Rack::Session::Dalli.

mwpastore:
Hi @petergoldstein, am I putting the mutex back in or can we move forward with this as-is?

petergoldstein:
@mwpastore So I don't like it, but for the moment can you put the mutex back and confirm that the performance regression goes away?

mwpastore:
No problem. Should we enable the mutex only when `@pool` is a Dalli::Client, or in all cases?

mwpastore:
@petergoldstein I went the former route. I think we're good to go. I'd like to leave it (adding the mutex back in) as a separate commit to make it easier to revert at a later time, if it ever becomes possible. 

Anyway, here are the benchmark results:

rack-session rebased:

```
Running 1m test @ http://localhost:9292
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   394.41ms  166.44ms   1.36s    66.31%
    Req/Sec    82.88     27.11   220.00     65.19%
  59473 requests in 1.00m, 13.73MB read
Requests/sec:    989.82
Transfer/sec:    234.06KB
```

rack-session rebased with mutex:

```
Running 1m test @ http://localhost:9292
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   353.84ms   99.22ms 927.41ms   74.01%
    Req/Sec    92.46     29.26   280.00     71.71%
  66287 requests in 1.00m, 15.31MB read
Requests/sec:   1102.99
Transfer/sec:    260.83KB
```

master (current):

```
Running 1m test @ http://localhost:9292
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   367.59ms  118.25ms   1.25s    70.30%
    Req/Sec    88.94     29.54   260.00     70.49%
  63860 requests in 1.00m, 14.75MB read
Requests/sec:   1062.63
Transfer/sec:    251.30KB
```

petergoldstein:
@mwpastore I know this has taken forever to get in, and my apologies for that.  Will be taking a final look this week, with a plan to get it into a release this upcoming weekend.

mwpastore:
No worries. Thanks @petergoldstein.

petergoldstein:
Long overdue review of this.  Looks good - merging to master.

mwpastore:
Thanks Peter!

PR #606, opened 2016-02-14
Execute tests for Rails 5

yui-knk:
Current .travis.yml does not include 'gemfiles/rails5.gemfile' to
gemfile, so tests for Rails 5 are not executed.


petergoldstein:
Thanks for catching that

PR #612, opened 2016-04-13
Add an option to raise an exception when trying to save too much data

himdel:
When the payload to `set` is larger than `value_max_bytes`, it's currently almost quietly ignored (logs a warning), appearing to have succeeded.

This adds an `fatal_over_max` option (defaulting to false - the previous behaviour), that causes `guard_max_value` to raise a `Dalli::TooBig` exception in such cases, making the failure more explicit and noticeable.

(test failures seem unrelated)

Thoughts, anyone? :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,8 @@ If serving compressed data using nginx's HttpMemcachedModule, set `memcached_gzi
 
 **value_max_bytes**: The maximum size of a value in memcached.  Defaults to 1MB, this can be increased with memcached's -I parameter.  You must also configure Dalli to allow the larger size here.
 
+**error_when_over_max_size**: Boolean. If true, Dalli will throw a Dalli::ValueOverMaxSize exception when trying to store data larger than **value_max_bytes**. Defaults to false, meaning only a warning is logged.
+
 **username**: The username to use for authenticating this client instance against a SASL-enabled memcached server.  Heroku users should not need to use this normally.
 
 **password**: The password to use for authenticating this client instance against a SASL-enabled memcached server.  Heroku users should not need to use this normally.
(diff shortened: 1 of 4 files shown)


petergoldstein:
I'm not exactly sure why Dalli doesn't throw an error when the value exceeds a max size by default.  But given the constraint that we don't want to change existing behavior, I like the idea behind this PR.

That said, a few quick comments:

1. The name of the config parameter 'fatal_over_max' seems inappropriate.   Something like 'error_when_over_max_size' would me more accurate
2. A similar comment about the error name.  'TooBig' is too generic.  Something more descriptive like 'ValueOverMaxSize' would be clearer.
3. There need to be some specs included with this PR before I could consider merging it.

himdel:
@petergoldstein great, thanks! :)

Yeah, naming is hard, and your names do sound more descriptive, I'll update the PR with those names, aand add specs.

petergoldstein:
@himdel Any progress here?

himdel:
@petergoldstein thanks for reminding me! I've renamed the option and exception as you asked, and added specs for `guard_max_value` in `test/test_server.rb`.

petergoldstein:
Thanks!

tecnobrat:
Any idea when this will make it into a release :)

petergoldstein:
@tecnobrat Not sure.  I think it's safe to say I'll be able to do a release by mid-June sometime.  May very well be earlier

petergoldstein:
Looks like it will be more like late June - shooting for this next weekend.

tjschuck:
@petergoldstein Friendly (bump) for cutting a release ❤️

roman-zaharenkov:
Guys, do you have any plans to release this feature? It has been merge a year ago but no version has been released since that time.

petergoldstein:
@roman-zaharenkov Not quite a year, but a very fair complaint.  I have been meaning to do the work to survey and release a 2.7.x for some time.  My apologies.

I'll try and make the time this weekend.  My overall goal is to make this the final release of the 2.7.x series, and branch to 3.0.x.

PR #618, opened 2016-04-28
add 'expire_after' in the session, so it is configurable in the app

chubako:
if 'rack.session.options' is set in the app, it can be overriden by the defaults if a middleware returns between the session and the app

### DIFF
--- a/lib/rack/session/dalli.rb
+++ b/lib/rack/session/dalli.rb
@@ -55,7 +55,7 @@ module Rack
 
       def set_session(env, session_id, new_session, options)
         return false unless session_id
-        expiry = options[:expire_after]
+        expiry = new_session['expire_after'] || options[:expire_after]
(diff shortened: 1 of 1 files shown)


lib/rack/session/dalli.rb
```diff
@@ -55,7 +55,7 @@ def get_session(env, sid)
 
       def set_session(env, session_id, new_session, options)
         return false unless session_id
-        expiry = options[:expire_after]
+        expiry = new_session['expire_after'] || options[:expire_after]
```
petergoldstein: Why is the key for `new_session` a string, while the key for `options` is a symbol?

chubako:
"options" is a Hash that passes options from the middleware configuration, so it's natural to use symbols as keys. "new_session" is the session data as a Hash. Session data is serialized for memcache store, so it's natural to use strings as keys.

PR #619, opened 2016-05-05
Change hash syntax to pre 2.0.0

brazir:
Changing the hash syntax to pre 2.0.0 will allow this version of the lib to be used in older ruby versions.

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -74,7 +74,7 @@ module Dalli
       end
     end
 
-    CACHE_NILS = {cache_nils: true}.freeze
+    CACHE_NILS = {:cache_nils => true}.freeze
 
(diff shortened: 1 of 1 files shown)


petergoldstein:
This is not 2.0.0 hash syntax, it's Ruby 1.9 hash syntax.  And Dalli only supports Ruby 1.9.3+.  What precise Ruby version are you trying to use with Dalli?

brazir:
I am on 1.8.7 right now, but if the syntax is supported by 1.9.3 then you can close this commit.  I was going based on the docs where they only show the new syntax is being used in the 2.0.0 version.
http://ruby-doc.org/core-1.9.3/Hash.html
http://ruby-doc.org/core-2.0.0/Hash.html

PR #622, opened 2016-05-25
global config raw options

yangxing-star:


petergoldstein:
As the globally failing tests indicate, this approach needs some work.  `options` is nil by default.  I'd also need you to add some specs to consider the PR for merging.

Thanks!

test/test_rack_session.rb
```diff
@@ -252,10 +252,10 @@
 
     res0 = req.get("/")
     session_id = (cookie = res0["Set-Cookie"])[session_match, 1]
-    ses0 = rsd.pool.get(session_id, true)
+    ses0 = rsd.pool.get(session_id, cache_nils: true)
```
petergoldstein: @yangxing-star How are the changes in test_rack_session.rb relevant to the PR?  I don't understand why this was changed from `true` to `{ cache_nils: true }`

PR #626, opened 2016-07-19
Add ValueOverMaxSize to the whitelist of expected exceptions that get re-raised

tjschuck:
This is a follow-up to #612, which doesn't currently behave as expected.

When actually using the `Server` instance as intended through the `request` interface, the `Dalli::ValueOverMaxSize` error was getting swallowed by the final `rescue => ex` branch since it wasn't added to the cases of whitelisted exceptions to re-raise.

Because of the way `test_server.rb` is written (using `send` to test private methods directly), there's unfortunately not really a good way to fix the tests to red/green this without serious refactoring.


petergoldstein:
Thanks for circling back on this and resolving the issue.

tjschuck:
@petergoldstein Thank you! :heart: :heart: :heart_eyes_cat: :heart:

Issue #573, opened 2015-11-30
Eliminate ActiveSupport::Cache::DalliStore and use ActiveSupport::Cache::MemCacheStore


petergoldstein:
Now that Rails is using Dalli by default for ActiveSupport::Cache::MemCacheStore, it makes sense to mark ActiveSupport::Cache::DalliStore deprecated with the next major release of Dalli, and direct people to use the packaged MemCacheStore. 

There doesn't appear to be much difference between the two backend classes at this point.  We can address any existing behavioral differences with the next major release of Dalli, then remove the DalliStore with the following major release.

petergoldstein:
Directly relates to issue #557 .

PR #634, opened 2016-10-04
Adding support for Rails5 CacheStore

austinrfnd:
Adding support for Rails 5 CacheStore but still sense for Rails3,4 AbstractStore.


wjordan:
cf. #573 this class may soon be deprecated, I'd suggest migrating `ActiveSupport::Cache::MemCacheStore` instead for Rails compatibility moving forward.

shadoath:
Bump for Rails 5

Issue #602, opened 2016-02-13
Trim Rails, Ruby support to recent versions


petergoldstein:
As part of the 3.0 release, it would be good to trim the support matrix a bit.

For Rails, I'd like to drop support for Rails 3.x at a minimum - limiting support to Rails 4+.  Ideally we'd drop support for anything older than 4.2.x, but that may be too restrictive.

Similarly, I'd like to trim the Ruby matrix.  Specifically:
- Drop support for Rubinius 2.x
- Drop support for JRuby before JRuby 9000
- Drop support for older MRI versions, potentially all versions prior to 2.2.  A more inclusive set may be appropriate here, but at a minimum I'd like to drop 1.9.3

mperham:
#758

PR #635, opened 2016-10-29
Use Ruby 2.3 methods for consistent hash ring

martijn:
Since #602 mentions dropping compatibility with old Ruby versions in Dalli 3.0 I thought it would be nice to use some of the new built-in tricks.

This changes replaces the current Dalli::Ring#binary_search method with Ruby's Array#bsearch_index method whilst maintaing compatible hashing. Performance improves 15% compared to the old Ruby implementation. I haven't benchmarked the inline C version, but I wonder how often that's even used considering the undocumented nature of that feature.

### DIFF
--- a/lib/dalli/ring.rb
+++ b/lib/dalli/ring.rb
@@ -21,7 +21,7 @@ module Dalli
             continuum << Dalli::Ring::Entry.new(value, server)
           end
         end
-        @continuum = continuum.sort { |a, b| a.value <=> b.value }
+        @continuum = continuum.sort_by(&:value)
       end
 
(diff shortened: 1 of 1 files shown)


martijn:
That's actually a 15% improvement on 10.000 get operations, so it's quite significant.

Didn't add a change log entry since there doesn't really seem to be a 'head' in there.

petergoldstein:
@martijn Agreed.  This is super interesting.  I'm booked solid for the next week or two, but I'm hoping to kick off the Dalli 3.x version shortly thereafter.

PR #641, opened 2016-12-01
Allow fetch_multi to use a splay of hashes to enable cache setting beyond using key

eko-vs:
# Description
Allow fetch_multi to use a splay of hashes to enable cache setting beyond using key

# Risks
This is NOT backwards comptabily with projects using hashes as keys

# TO DO
- [ ] Add comments and examples for use


petergoldstein:
I don't really see this as a general use case.  In this circumstance I'm unclear why you simply wouldn't project out the :key value of the Hash key before calling read_multi.  And as there still aren't any comments/examples of use, I'm going to close this PR.

If you come up with some compelling use cases, please feel free to share here and I'll consider reopening.  Thanks.

PR #643, opened 2016-12-26
Keep @ring object of Dalli::Client after #close to reuse it on next operation

key-amb:
Hello,

In my environment, `Dalli::Ring.new` was a bottleneck for performance.
So I want to reuse `@ring` objects even after `Dalli::Client#close` (or `#reset`) is called.
I already made a monkey-patch in my environment and it works well.

But this change can be a breaking change and will affect so many people.
Second choice is here: https://github.com/petergoldstein/dalli/compare/master...key-amb:topic/reuse-ring
This just makes `Dalli::Client#ring` public so that we can access `@ring` from application codes.

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -250,10 +250,20 @@ module Dalli
     ##
     # Close our connection to each server.
     # If you perform another operation after this, the connections will be re-established.
-    def close
+    def close!
       if @ring
         @ring.servers.each { |s| s.close }
         @ring = nil
+      end
+    end
+    alias_method :reset!, :close!
+
+    ##
+    # Close our connection to each server like +#close!+ but do not clear ring
+    # so that one can reuse it for next operation.
+    def close
+      if @ring
+        @ring.servers.each { |s| s.close }
       end
     end
     alias_method :reset, :close


petergoldstein:
@key-amb I'm not sure this is a general use-case problem.

How was Dalli::Ring a bottleneck on performance if it's only cleared after closing the Dalli::Client?  How often were you closing the clients?   Were you using a connection pool of Dalli::Client objects?

key-amb:
Hello,

Thank you for your reply.
Maybe this PR is not for general use because my environment is kind of big.

> Were you using a connection pool of Dalli::Client objects?

I don't use any connection pool because I run over 20K web application processes
and memcached servers are difficult to keep all connections for all web-app processes.

> How often were you closing the clients? 

Before I implemented the monkey-patch, each web-app process had called `Dalli::Client#close`
every after it responded to a request.  
Each process handles less than 10 requests per second.

> How was Dalli::Ring a bottleneck on performance if it's only cleared after closing the Dalli::Client? 

I think I had called `Dalli::Client#close` so often that it made `Dalli::Ring#new` a
bottleneck as a result.

petergoldstein:
Ok, that would explain it.  I don't think that's a general use case - opening and closing clients per request isn't a supported pattern.  So I'm going to close this.

key-amb:
I understand. Thank you for taking time!

PR #645, opened 2017-01-04
ActiveSupport instrument DalliStore errors

btatnall:
Previously, one could use raise_errors to either handle execptions
in client code or have them handled by dalli. In the later case,
logging was the only way to be notified of an exception. This
commit adds the ability to be notified via active support
instrumentation via a cache_error.active_support event.

I've added this code because I want to know when errors are
occuring to trigger external monitoring and I still want
dalli to fail without raising errors so that cached values
are still generated.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,8 @@
 Dalli Changelog
 =====================
 
+- Instrument DalliStore errors with instrument_errors configuration option. (btatnall)
+
 2.7.6
 ==========
 - Rails 5.0.0.beta2 compatibility (yui-knk, petergoldstein)
--- a/README.md
+++ b/README.md
@@ -194,6 +194,10 @@ If serving compressed data using nginx's HttpMemcachedModule, set `memcached_gzi
 
 **cache_nils**: Boolean. If true Dalli will not treat cached `nil` values as 'not found' for `#fetch` operations. Default is false.
 
+**raise_errors**: Boolean. When true DalliStore will reraise Dalli:DalliError instead swallowing the error. Default is false.
+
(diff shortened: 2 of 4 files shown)


btatnall:
I didn't add any documentation to the readme because I didn't know if you would want this configuration option to be public in the same vein as raise_errors.

btatnall:
It looks like bundler install failed for the '1 errored check'.

> The command "bundle install --jobs=3 --retry=3 --path=${BUNDLE_PATH:-vendor/bundle}" failed and exited with 5 during .

I don't believe this has anything to do with the file changes that I made.

petergoldstein:
@btatnall I agree - looks like it's a Gemfile issue, because the latest Nokogiri no longer supports Ruby 2.0.x.  At first pass this looks good - I'll take a closer look this weekend and merge/reply as needed.

Per your documentation question, I'd expect this option to be public in the same vein as raise_errors.  So if you haven't already added such documentation as part of the PR, please do.

btatnall:
@petergoldstein raise_errors is **NOT** a public option and it isn't documented. I can add documentation for that as well if it should be public. Doesn't matter either way to me.

petergoldstein:
@btatnall Looking back at the documentation history it looks like raise_errors should have been included in the options, but was added to the Rails 3.x section instead, and removed when that section was revised.

If you could please add raise_errors and the new option to the README.md, that would be helpful.

btatnall:
@petergoldstein I've added the two errors configuration options to the README.md.

btatnall:
@petergoldstein Is it possible to get a release version that includes this merge? Thanks!

Issue #621, opened 2016-05-23
down_retry_delay default is too low


mrsimo:
Hi there! We were running some tests to figure out how much memcached downtime could handle our app, and since that downtime, we've gotten a bit familiar with `down_retry_delay`.

The default value for this value is 1s, but coupled with a `socket_max_failures` of 2, and a `socket_timeout` of 0.5, I think that makes the default for `down_retry_delay` not very good.

If the memcached server is down in a way that we exhaust the `socket_timeout`, considering the retries it does, `down_retry_delay` doesn't ever do anything.

Stopping one of our memcached servers caused all memcached queries to it to take 1s, which slowed our unicorns to a point where we ran out of them.

We changed the value of `down_retry_delay` to 120s, and now memcached servers going down doesn't have any visible effects.

I would suggest changing the default of this value to something bigger.

petergoldstein:
Fixed by #650

PR #650, opened 2017-03-08
Update down_retry_delay default

jaredhales:
Update `down_retry_delay` for #621

I ran into this myself recently. I saw the existing issue and figured I’d throw together a quick pull request since nobody had done anything with this since it was raised back in May of last year.


petergoldstein:
Thanks!

PR #654, opened 2017-05-01
Add details of port and weight configuration to documentation.

dkam:
Hi.  I've updated some documentation to include references to setting the server port and weight.


petergoldstein:
Thanks!

PR #661, opened 2017-08-22
Use `assert_nil` in tests

pvalena:
instead of `assert_equal nil`, because it produces errors with activesupport >= 5.1.3.


petergoldstein:
@pvalena Thanks for the contribution.  Would up inadvertently duplicating this in PR #665 .  So I'm closing this PR.

PR #662, opened 2017-08-29
Not found checks no longer trigger the result's equality method

dannyfallon:
Triggering the equality method on the result from the fetch can
cause the state of the result object to change, for example if it has
a custom `==` instance method. By reversing the operands we end up
achieving a hit/miss determination but will call `#==` on `nil` or
on `Dalli::Server::NOT_FOUND` if `cache_nils` is `true`. Either way
we won't modify the returned object from the cache before sending it
back to the caller.

### DIFF
--- a/lib/active_support/cache/dalli_store.rb
+++ b/lib/active_support/cache/dalli_store.rb
@@ -103,13 +103,13 @@ module ActiveSupport
               read_entry(namespaced_name, options).tap do |result|
                 if payload
                   payload[:super_operation] = :fetch
-                  payload[:hit] = result != not_found
+                  payload[:hit] = not_found != result
                 end
               end
             end
           end
 
-          if entry == not_found
+          if not_found == entry
             result = instrument_with_log(:generate, namespaced_name, options) do |payload|
               yield
             end
--- a/test/test_active_support.rb
+++ b/test/test_active_support.rb
@@ -9,6 +9,12 @@ class MockUser
   end
 end
 
+class ObjectRaisingEquality
+  def ==(other)
+    raise "Equality called on fetched object."
+  end
+end
+
 describe 'ActiveSupport::Cache::DalliStore' do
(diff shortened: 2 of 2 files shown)


dannyfallon:
My goal here is to ensure that `result` is unmodified as a result of the instrumentation. 

My specific example type of object that will potentially change state if you call `==` on it is an `ActiveRecord::AssociationRelation` object. We cached this thinking that the serialiser would load the records and we would end up caching an array of `ActiveRecord` objects but we just cached the association. On fetch the instrumentation comparison for a hit/miss forced us to load all the records whcih I think is kinda weird. I totally understand that he use of the cached object would probably have done the same thing and loaded the records but we've better visibility into that than having it done by gem action at a distance IMO

petergoldstein:
@dannyfallon So I get the goal, although this is a super-edgy case.  And since it isn't a change in behavior and it's a very small code change, I'm inclined to merge it.

But as is, it's almost certain to get reverted at some point in the future.  You've added a test case, but there's no corresponding comment as to why '==' shouldn't be called.  And since this is so edgy, that reasoning is not obvious.

I'd suggest you add a comment - probably in the spec - as to the motivation for this change, possibly with a link to this PR.

dannyfallon:
@petergoldstein, thanks for the review and for the comments. I've made a number of small changes:

1. Swapped to using a class that raises if `#==` is called, rather than using a mock
2. Moved the test outside of the `instruments` block since there is an equality check in `#fetch` that determines if we should write the value to the cache.
3. Added a comment to link to the PR for more context on why this may be a bad idea.

petergoldstein:
@dannyfallon Can you please rebase against master?

dannyfallon:
Absolutely, I'll do it as soon as I'm near the laptop :+1:

petergoldstein:
Thanks @dannyfallon

PR #663, opened 2017-08-29
micro cleanup

grosser:
- use sort_by since it is easier to read and faster for large lists 
 - remove unnecessary assignment and return


petergoldstein:
Closing in favor of #677

Issue #673, opened 2018-02-16
Create a docker image to run the suite


sobrinho:
Just to keep the things open and explicit, I've talked with @petergoldstein and we are going to create a Docker image to run the suite in Travis and take care of the open issues on Dalli.

@tiarly offered to create the Docker image then we will wait for that to fix the Travis build in sequence.

sobrinho:
We managed this on #674.

PR #674, opened 2018-02-23
Fix travis memcached-sasl installation

tiarly:
The idea here was to build an actual docker build, but since we're kinda way too blocked on making our travis build to work this seemed to be a faster way.

In response to https://github.com/petergoldstein/dalli/issues/673


petergoldstein:
@tiarly - Thanks so much.  I merged your changes, along with a few other modifications necessary to get the build running in PR #675

PR #679, opened 2018-03-16
Don't override generate_sid

pixeltrix:
There's no need to override the generate_sid method and call the super method via an alias because the get_session and destroy_session methods call generate_sid_with directly. This fixes the issue with the Rails subclass of Rack::Session::Dalli which overrides the generate_sid method for compatibility reasons.

### DIFF
--- a/lib/rack/session/dalli.rb
+++ b/lib/rack/session/dalli.rb
@@ -157,18 +157,9 @@ module Rack
         [nil, mserv, mopts, popts]
       end
 
-      # Capture generate_sid's super so we can call it from generate_sid_with
-      alias_method :generate_sid_super, :generate_sid
-
-      def generate_sid
-        # no way to check env['rack.multithread'] here so fall back on
-        # Dalli::Client or ConnectionPool's internal mutex cf. our own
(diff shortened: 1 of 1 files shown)


petergoldstein:
@pixeltrix When looked at from the larger perspective I'm not sure this is the right fix.  The ultimate goal should be to make DalliStore and MemCacheStore converge - there's really no reason that I can see to have two separate classes.

But because MemCacheStore does override generate_sid (through the Compatibility module), this change doesn't serve that larger goal.

Thoughts?

pixeltrix:
The only differences in `ActionDispatch::Session::MemCacheStore` are to do with Rails compatibility so to just use `Rack::Session::Dalli` directly would require us to break that. Similarly including the Rails specific things in `Rack::Session::Dalli` would break compatibility with existing Rack apps.

Since this is a patch release I don't think either of those are good options - we can investigate whether we can unify things for Rails 6.0 and Dalli 3.0.

petergoldstein:
@pixeltrix Ok, I'll buy that.  Seems like a reasonable approach.

PR #682, opened 2018-03-26
Introduce new option, instrument_namespace, when running multiple dalli stores

shayonj:
Proposal/Feature request PR :).

When we have more than one dalli store implementation in an app,
this option config would allow to subscribe to the same
`ActiveSupport::Notifications` operation(s), but also allow the ability to
segment these through an additional payload key `instrument_namespace`.
Otherwise, we could have the same instrumentation metrics
(`cache_read.active_support`) from different dalli store implementations
and no way to differentiate between  them.

### DIFF
--- a/lib/active_support/cache/dalli_store.rb
+++ b/lib/active_support/cache/dalli_store.rb
@@ -388,6 +388,8 @@ module ActiveSupport
       end
 
       def instrument(operation, payload)
+        payload[:instrument_namespace] = @options[:instrument_namespace] if @options[:instrument_namespace]
+
         ActiveSupport::Notifications.instrument("cache_#{operation}.active_support", payload) do
           yield(payload) if block_given?
         end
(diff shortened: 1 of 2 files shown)


shayonj:
@petergoldstein would love to know if you think this might be a good addition? :)

shayonj:
cc @petergoldstein would love to know your thoughts on this :)

petergoldstein:
@shayonj Thanks for submitting, and sorry it's taken me so long to reply.

I'm not averse to merging something like this, but I'd like to be as standard as possible.  Is 'instrument_namespace' a standard name for this kind of key?  If it isn't, is there a more standard one?  What do other systems use in this case?

shayonj:
Alternatively, we can also call it `identifier` / `id`. Concise.

mperham:
dalli_store has been removed.

PR #696, opened 2018-08-28
Reduce array allocations

schneems:
In the case where a method is called on an array that returns a new array, we can mutate that new array instead of generating more intermediate arrays.

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -64,7 +64,10 @@ module Dalli
     # If a block is given, yields key/value pairs one at a time.
     # Otherwise returns a hash of { 'key' => 'value', 'key2' => 'value1' }
     def get_multi(*keys)
-      return {} if keys.flatten.compact.empty?
+      check_keys = keys.flatten
+      check_keys.compact!
+
+      return {} if check_keys.empty?
       if block_given?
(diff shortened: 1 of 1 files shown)


petergoldstein:
Thanks!

lib/dalli/client.rb
```diff
@@ -292,7 +295,9 @@ def groups_for_keys(*keys)
     end
 
     def mapped_keys(keys)
-      keys.flatten.map {|a| validate_key(a.to_s)}
+      keys_array = keys.flatten
+      keys_array.map! { |a| validate_key(a.to_s) }
+      keys_array
```
avokhmin: why need to use `keys_array `?
`#map!` returns the same:

```
> x = [1, 2, 3]
 => [1, 2, 3] 
> x.map! { |i| i * 2  }
 => [2, 4, 6] 
```

schneems: [].map! returns nil

avokhmin: screen shot 2018-10-24 at 10 41 11

hm, strange. In which version of ruby?

schneems: I was on mobile so I ran it in the browser. Guess it has a different behavior. Ignore my previous comment.

PR #697, opened 2018-08-30
Faster expanded_key

schneems:
- Only check if a key responds to `cache_key` if it is not an array, hash, or string.
- Only duplicate the key when it’s a string (a more elegant way to than recording object_id.
- Move `to_param` logic for arrays to this file. Calling `to_param` on an array generates a new array. We can save that extra array allocation by reproducing the logic here
- Mutate array generated by hash instead of creating new temporary arrays.
- Reproduce to_param logic for array in the to_hash.

### DIFF
--- a/lib/active_support/cache/dalli_store.rb
+++ b/lib/active_support/cache/dalli_store.rb
@@ -350,25 +350,26 @@ module ActiveSupport
       # object responds to +cache_key+. Otherwise, to_param method will be
       # called. If the key is a Hash, then keys will be sorted alphabetically.
       def expanded_key(key) # :nodoc:
-        return key.cache_key.to_s if key.respond_to?(:cache_key)
-        original_object_id = key.object_id
-
         case key
         when Array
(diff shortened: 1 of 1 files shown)


schneems:
Also feel free to let me know if you don't like one or more of these optimizations.

On a sidenote I was profiling memcached with dalli versus redis caching in rails in codetriage and by my very unscientific tests it looks like dalli and memcached were both faster and allocated less memory in Ruby land. Thanks a ton for this gem.

petergoldstein:
@schneems Can you please rebase and incorporate the cache_key_with_versioning fix from master?  Under those conditions I think it makes sense to merge.  Thanks!

schneems:
I actually think we should close this. I'm doing work in Rails which which will make this patch not really needed https://github.com/rails/rails/pull/33944

It adds a Key object that pre-generates a `cache_key` value that would get passed to `expanded_key` here and immediately return.

PR #698, opened 2018-09-07
rails 5.2 cache_versioning support

GriwMF:
rails 5.2 introduced Recyclable cache keys https://github.com/rails/rails/pull/29092
which means that with Rails.application.config.active_record.cache_versioning = true dalli does not work (#cache_key now returns key without version. We need to use #cache_key_with_version)

### DIFF
--- a/lib/active_support/cache/dalli_store.rb
+++ b/lib/active_support/cache/dalli_store.rb
@@ -346,10 +346,12 @@ module ActiveSupport
       end
       alias :normalize_key :namespaced_key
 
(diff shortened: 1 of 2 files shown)


jeremy:
CI failure is unrelated to this change, btw.

petergoldstein:
Thanks, I just caught up on this.  I've resolved the CI failure and I'm going to merge this.  I'm also planning to release a gem version with this PR included shortly.  Thanks @GriwMF !

PR #699, opened 2018-09-12
Show warning message if try to serialize non basic data type objects

hlcfan:
I happen to meet this issue while upgrading Rails from 3 to 4.

Before upgrade, we have:
```
Dalli::Client.new(host, namespace: "namespace", compress: false, serializer: JSON)
```

After upgrade, it stores "weird" data to Memcache, like `"\"#<User:0x007f83dc017248>\""`, instead of `"{\"id\":1,\"name\":\"al\"}"`.

Rails changes are:
+ https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#json-gem-compatibility
+ https://github.com/rails/rails/blob/3-2-stable/activesupport/lib/active_support/core_ext/object/to_json.rb#L15
+ https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/object/json.rb#L36

I think it'd be better to show a warning message when try to cache non basic data type objects.
BTW, different kinds of JSON serializers have different behaviors. Some of them:

```
struct = Struct.new(:name).new("alex")
JSON.dump(struct) # "\"#<struct name=\\\"alex\\\">\""
Yajl.dump(struct) # "\"#<struct name=\\\"alex\\\">\""
Oj.dump(struct)   # "{\"^u\":[[\"name\"],\"alex\"]}"
```

### DIFF
--- a/lib/dalli/server.rb
+++ b/lib/dalli/server.rb
@@ -204,6 +204,15 @@ module Dalli
 
     private
 
+    SKIP_WARNING_DATA_TYPES = [Array, FalseClass, TrueClass, Hash, Float, Fixnum, NilClass, String, Symbol]
+    private_constant :SKIP_WARNING_DATA_TYPES
+
+    def warn_unless_value_is_basic_data_type value
+      if [value].flatten.detect { |v| !SKIP_WARNING_DATA_TYPES.include?(v.class) }
+        Dalli.logger.warn("You're serializing a Ruby Class object which may not be serialized properly. Please convert to basic data type first.")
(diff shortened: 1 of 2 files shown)


petergoldstein:
@hlcfan Thanks for the PR, but I don't think this would be a good addition to Dalli.  Specifically:

1. It would add an unnecessary warning in lots of success cases (when the object being inserted is serializable, but not one of the types you're detecting
2. It adds overhead to all insertions

So I'm going to close this PR without merging.  Thanks.

PR #700, opened 2018-09-13
Update README for cache_versioning in Rails 5.2

brchristian:
This PR adds some helpful information for developers using Dalli in Rails 5.2 and later. Rails 5.2 changed the way that cache keys function, and until cache versioning is supported in the current version of Dalli, users will need to make sure that `cache_versioning` is set to `false` in their application config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,16 @@ Rails.application.config.session_store :dalli_store, :memcache_server => ['host1
 Dalli does not support Rails 2.x.
 
 
+Usage with Rails 5.2 and 6.x
+---------------------------
+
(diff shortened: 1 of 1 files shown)


schneems:
Does cache invalidation work with that setting turned off? https://github.com/rails/rails/issues/33907

NARKOZ:
In rails you can use mem_cache_store which supports recyclable keys and under the hood uses dalli.

petergoldstein:
PR #698 is merged to master, which addresses the `cache_versioning` issue.  I'm going to close this PR without merging.  I should have a new gem version sometime in the next couple of days.

Thanks for contributing!

PR #701, opened 2018-09-28
[BUG?] #fetch does not provide a key attribute to the fallback block (but should provide)

0exp:
`ActiveSupport::Cache::Storage` abstraction requires that `#fetch`/`#fetch_multi` methods provides a key to the block used for generating entries that does not exist in cache storage (with a given key).

I found that `ActiveSupport::Cache::DalliStore#fetch` missed this logic (in my own project).

This pull request resolves this.

now it works correctly:

```ruby
# INCORRECT:
cache.fetch("a") { |key| "#{key}-calculated" } # in current implementation the key is nil
# => { "a" => "-calculated" }

# CORRECT (pull request)
cache.fetch("a") { |key| "#{key}-calculated" } # in current pull request the key has a corresponding value
# => { "a" => "a-calculated" }

### DIFF
--- a/lib/active_support/cache/dalli_store.rb
+++ b/lib/active_support/cache/dalli_store.rb
@@ -111,7 +111,7 @@ module ActiveSupport
 
           if not_found == entry
             result = instrument_with_log(:generate, namespaced_name, options) do |payload|
-              yield
+              yield(name)
             end
             write(name, result, options)
             result
--- a/test/test_active_support.rb
+++ b/test/test_active_support.rb
@@ -77,6 +77,14 @@ describe 'ActiveSupport::Cache::DalliStore' do
         @dalli.fetch('obj') { obj }
       end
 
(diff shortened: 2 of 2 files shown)


petergoldstein:
@0exp Can you please point me to the reference in the original abstraction?  I'm not seeing this in the docs.

0exp:
@petergoldstein sure:

#fetch_multi provides a cache key name:
https://github.com/rails/rails/blob/master/activesupport/lib/active_support/cache.rb#L439

# fetch provide a cache key name:
https://github.com/rails/rails/blob/master/activesupport/lib/active_support/cache.rb#L331

0exp:
This is really useful when you work with a set of keys (via `#fetch_multi`)

petergoldstein:
@0exp Interesting.  Definitely there in the code, but not at all reflected in the docs.  All of their example blocks have no parameters.

Thanks for the contribution.

Issue #704, opened 2018-10-24
Seemingly version 2.7.9 has problems with frozen strings


kikonen:
Appeared after updating from 2.7.8 to 2.7.9 version.

```
ERROR [2018-10-24T05:00:51.777778 #14063-22175620]  #<ActionView::Template::Error: can't modify frozen String>
xxx/shared/bundle/ruby/2.5.0/gems/dalli-2.7.9/lib/active_support/cache/dalli_store.rb:372:in `force_encoding'
xxx/shared/bundle/ruby/2.5.0/gems/dalli-2.7.9/lib/active_support/cache/dalli_store.rb:372:in `expanded_key'
xxx/shared/bundle/ruby/2.5.0/gems/dalli-2.7.9/lib/active_support/cache/dalli_store.rb:361:in `block in expanded_key'
xxx/shared/bundle/ruby/2.5.0/gems/dalli-2.7.9/lib/active_support/cache/dalli_store.rb:361:in `collect'
xxx/shared/bundle/ruby/2.5.0/gems/dalli-2.7.9/lib/active_support/cache/dalli_store.rb:361:in `expanded_key'
xxx/shared/bundle/ruby/2.5.0/gems/dalli-2.7.9/lib/active_support/cache/dalli_store.rb:340:in `namespaced_key'
xxx/shared/bundle/ruby/2.5.0/gems/dalli-2.7.9/lib/active_support/cache/dalli_store.rb:129:in `read'
xxx/shared/bundle/ruby/2.5.0/gems/actionpack-5.2.1/lib/abstract_controller/caching/fragments.rb:114:in `block in read_fragment'
xxx/shared/bundle/ruby/2.5.0/gems/actionpack-5.2.1/lib/abstract_controller/caching/fragments.rb:162:in `block in instrument_fragment_cache'
xxx/shared/bundle/ruby/2.5.0/gems/activesupport-5.2.1/lib/active_support/notifications.rb:168:in `block in instrument'
xxx/shared/bundle/ruby/2.5.0/gems/activesupport-5.2.1/lib/active_support/notifications/instrumenter.rb:23:in `instrument'
xxx/shared/bundle/ruby/2.5.0/gems/activesupport-5.2.1/lib/active_support/notifications.rb:168:in `instrument'
xxx/shared/bundle/ruby/2.5.0/gems/actionpack-5.2.1/lib/abstract_controller/caching/fragments.rb:162:in `instrument_fragment_cache'
[...]

PR #705, opened 2018-10-24
Test supports frozen strings

schneems:


petergoldstein:
@schneems How do you square this passing spec with the bug reports in #704 ?

schneems:
I think we need more info from them. I’ll ask on the ticket if they can give us more info about the error or what keys they’re passing in.

schneems:
I’m fine if you want to revert my performance patch while we are trying to figure out a reproduction.

petergoldstein:
Closing because this test didn't actually test the issue.

PR #706, opened 2018-10-24
Advertise support for "cache versioning"

schneems:
Introduced in https://github.com/rails/rails/pull/33943 Rails will attempt to interpret if a cache store supports "cache versioning" by looking for the predicate method `supports_cache_versioning?` on the cache store class. This PR advertises support that was introduced in https://github.com/petergoldstein/dalli/pull/698.


petergoldstein:
@schneems Thanks.

PR #721, opened 2019-03-18
Remove kgio support

deepj:
kgio is not recommended to use with Ruby 2.3 and later https://bogomips.org/kgio/

I removed also `Performance.md` which is out-of-date and pretty old. The recent version of dalli seems support only Rails 5+ so I removed `rails4.gemfile` in this PR.

Rails 5.0 supports Ruby 2.2 but this versions is almost over 1 year in EOL. https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/

I guess, there is a time to remove kgio from dalli source.


mperham:
Socket IO has been rewritten to work with Ruby 2.3+ and remove kgio.

PR #724, opened 2019-05-23
Added digest_class configuration option

### LINKED ISSUE #723: Dalli uses MD5 which is not permitted in FIPS mode
When providing services to United States federal agencies, you must implement FedRAMP controls, one of which requires using FIPS for cryptographic operations. In Rails 5.2 work was done to remove explicit calls to MD5 and replace them with the `ActiveSupport::Digest` class.

As there are comments in the change log indicating Rails 5.2 support https://github.com/petergoldstein/dalli/blob/master/History.md#278, I would expect that Dalli would also implement similar mechanisms as Rails did with their mem_cache_store adapter.

https://github.com/rails/rails/blob/v5.2.0/activesupport/lib/active_support/cache/mem_cache_store.rb#L188

Perhaps instead of fixing the `DalliStore` class, https://github.com/petergoldstein/dalli/issues/573 should be finally completed?

However, I do notice that the `Dalli::Client` class also calls MD5 https://github.com/petergoldstein/dalli/blob/master/lib/dalli/client.rb#L382, and given that Dalli is intended to be used independent of Rails, I'm not sure what the correct solution would be for that. Perhaps a configuration options can be set to pass in the desired Digest class? This way a default can be set and in Rails both methods will call `ActiveSupport::Digest`, but in non-Rails envs the client can use whatever Digest class the user prefers. Moreover, this solution would also support older versions of Rails that may not have `ActiveSupport::Digest`.

I am presently working on a PR to address this oversight. I hope that you can provide guidance on how you want to move forward with a solution that makes Dalli FIPS compatible. Thank you for your time.

saghaulor:
- MD5 is not permitted in FIPS mode, nor is the Digest class. When operating with FIPS mode, only OpenSSL::Digest::SHAxxx is permitted.
-Closes #723

### DIFF
--- a/lib/active_support/cache/dalli_store.rb
+++ b/lib/active_support/cache/dalli_store.rb
@@ -345,7 +345,7 @@ module ActiveSupport
         namespace = options[:namespace] if options
         prefix = namespace.is_a?(Proc) ? namespace.call : namespace
         key = "#{prefix}:#{key}" if prefix
-        key = "#{key[0, 213]}:md5:#{::Digest::MD5.hexdigest(key)}" if key && key.size > 250
+        key = "#{key[0, 213]}:md5:#{@options[:digest_class].hexdigest(key)}" if key && key.size > 250
         key
       end
       alias :normalize_key :namespaced_key
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -29,8 +29,10 @@ module Dalli
     # - :serializer - defaults to Marshal
     # - :compressor - defaults to zlib
     # - :cache_nils - defaults to false, if true Dalli will not treat cached nil values as 'not found' for #fetch operations.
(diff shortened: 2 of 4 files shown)


petergoldstein:
@saghaulor So a couple of pieces of feedback:

1. This change needs one or more corresponding specs
2. We should not just allow the passing in of a random class, especially since it may or may not implement the `hexdigest` method.  At a minimum there should be duck type checking.  Alternatively, we can specify a list of acceptable digest classes and make that configurable with a string value.

Happy to see either option on #2.

saghaulor:
@petergoldstein thanks for the great feedback. I'll add some more stringent checks on the digest_class option, and add some accompanying tests.

saghaulor:
@petergoldstein I believe this PR is ready to merge if you agree with the changes. Please let me know if there is anything else you would like me to do regarding this PR.

petergoldstein:
@saghaulor This looks good, thanks.  Will merge.  Not likely to get to a release immediately, but I'll try to do it within the next week or so.

PR #735, opened 2019-11-18
Add Rails 6 support

mrhead:
- Add a new appraisal for Rails 6.
- Run `appraisal generate` to generate new gemfiles.
- Set Rails 6 as a dev dependency.
- Test Rails 6 on Travis.


.travis.yml
```diff
@@ -7,7 +7,7 @@ rvm:
   - 2.3.8
   - jruby-9.1.16.0
 gemfile:
-  - gemfiles/rails5.gemfile
+  - gemfiles/rails6.gemfile
```
mrhead: @petergoldstein At the moment only Rails 5 is tested on Travis CI for the master branch so I specified only Rails 6 here. However I would suggest to test both, Rails 5 and Rails 6. What do you think about it?

mrhead: Actually I think that we should test both, Rails 5 and Rails 6. Please let me know if you think otherwise.

PR #736, opened 2019-11-18
Remove appraisal for Rails 3

mrhead:
A gemfile for Rails 3 doesn't exist in `gemfiles/` so I presume that Rails 3 appraisal isn't needed anymore.

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -1,7 +1,3 @@
-appraise 'rails3' do
-  gem 'rails', '>= 3.2.0', '< 4'
-end
-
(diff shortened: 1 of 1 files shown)


mrhead:
@petergoldstein What do you think about completely removing Rails 3 and Rails 4 support? At the moment Rails 3 and 4 are still mentioned in the README.

petergoldstein:
@mrhead My intention was to remove those with the long delayed 3.x release.  But given the time and status of those, I'd be good with you removing.

PR #748, opened 2020-05-16
Pass options hash to Client, Ring, and Server

mdehoog:
Adds support for passing an `options` Hash from any `Dalli::Client` calls into the downstream `Dalli::Client`, `Dalli::Ring` and `Dalli::Server` methods. This allows callers to pass options like `failover: false` in the cache call itself, rather than having to create a separate Client or Store to enable these options.

Normally this can be achieved by creating a separate Store to get different configuration, but right now there's no way to share connections between Stores. This means if you want a particular call to `failover` and another one to not, the only current option is to double the open connections (one pool for each configured Store).

Currently the `set`, `replace`, `get`, and `add` operations support passing `options`, but the `delete`, `incr`, `decr`, `append`, `prepend`, `cas`, and `touch` operations don't. This PR adds `options` support for these latter operations.

It also adds support for passing `:failover` option per operation, rather than as configuration.

An alternative direction would be to allow creation of new clients that share the same underlying pool of connections, but this is a deeper change as currently the connection pool is at the Store level, as a pool of Clients.

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -65,15 +65,16 @@ module Dalli
     # If a block is given, yields key/value pairs one at a time.
     # Otherwise returns a hash of { 'key' => 'value', 'key2' => 'value1' }
     def get_multi(*keys)
+      options = keys.last.is_a?(::Hash) ? keys.pop : {}
       keys.flatten!
       keys.compact!
 
       return {} if keys.empty?
       if block_given?
-        get_multi_yielder(keys) { |k, data| yield k, data.first }
+        get_multi_yielder(keys, options) { |k, data| yield k, data.first }
       else
         {}.tap do |hash|
-          get_multi_yielder(keys) { |k, data| hash[k] = data.first }
+          get_multi_yielder(keys, options) { |k, data| hash[k] = data.first }
         end
       end
     end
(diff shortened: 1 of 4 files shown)


mdehoog:
Would be interested to hear any thoughts on this proposal... we've been running this fork in production for a week and would love to get back on mainline at some point. Thanks!

mperham:
If you want to resubmit this PR, we'll take a look. Broadly, I'm not sure I like the idea of individual operations being able to override global settings but I could be convinced with a reasonable use case.

mdehoog:
Thanks @mperham, I've rebased the original change with the latest master.

The main benefit of having options passed is that we can do things like control the failover per operation.

As an example, our normal caching operations are failover tolerant in that they simply rely on the TTL, and it doesn't matter if a memcached node goes down and we temporarily store a value in another node in the ring.

However, we have certain operations that we want to prevent failover, as we prefer stronger cache invalidation guarantees. To achieve this we pass `failover: false` to all reads and writes to these keys, and `raise_errors: true` to writes. We also have a patch applied to `Dalli::Ring#server_for_key` method (irrelevant code omitted):
```ruby
module DalliRingWithServerOffset
  def server_for_key(key, options = nil)
...
        server = @continuum[entryidx].server
        server_offset = options[:server_offset].to_i
        server = @servers[(@servers.index(server) + server_offset) % @servers.length] if server_offset != 0
        return server if server.alive?
...
  end
end
Dalli::Ring.prepend(DalliRingWithServerOffset)
```

This allows us to store a key on two separate servers in the ring, which are both read and only used if equal, which gives us stronger cache key invalidation when one server is down.

We could do this with different combinations of options on the client itself, but we use quite a few different combinations of options:
 - `failover: true, raise_errors: false`
 - `failover: false, raise_errors: false, server_offset: 0`
 - `failover: false, raise_errors: false, server_offset: 1`
 - `failover: false, raise_errors: true, server_offset: 0`
 - `failover: false, raise_errors: true, server_offset: 1`

With the current gem, we'd require 5 different clients, and therefore 5 different connection pools. This is unfortunately untenable due to the large amount of connections to memcached, given Ruby's threading model.

mperham:
I think this is a little too specialized for Dalli. I’m happy to entertain patches or refactorings that make a fork or patch more maintainable

PR #749, opened 2020-05-19
DalliStore: Fix default digest_class

mdehoog:
#724 introduced the ability to pass in a `digest_class` option. It also introduced a bug into `DalliStore` where passing a key with a length greater than 250 without passing a `digest_class` option will result in a `NoMethodError: undefined method 'hexdigest' for nil:NilClass` error. This is because there's no fallback specified for this option in DalliStore.

This PR introduces this fallback, and also makes this fallback consistent in the `Dalli::Client` class (rather than mutating the incoming options hash, it uses a fallback in the caller). It also adds a regression test for the bug.

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -32,7 +32,6 @@ module Dalli
     # - :digest_class - defaults to Digest::MD5, allows you to pass in an object that responds to the hexdigest method, useful for injecting a FIPS compliant hash object.
     #
     def initialize(servers=nil, options={})
-      options[:digest_class] = ::Digest::MD5 unless options[:digest_class]
       @servers = normalize_servers(servers || ENV["MEMCACHE_SERVERS"] || '127.0.0.1:11211')
       @options = normalize_options(options)
       @ring = nil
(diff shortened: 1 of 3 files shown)


lib/dalli/client.rb
```diff
@@ -32,7 +32,6 @@ class Client
     # - :digest_class - defaults to Digest::MD5, allows you to pass in an object that responds to the hexdigest method, useful for injecting a FIPS compliant hash object.
     #
     def initialize(servers=nil, options={})
-      options[:digest_class] = ::Digest::MD5 unless options[:digest_class]
```
mdehoog, line 35: moving the fallback below avoids mutating the options parameter

lib/dalli/client.rb
```diff
@@ -353,9 +352,7 @@ def ring
             server_options[:password] = uri.password
             s = "#{uri.host}:#{uri.port}"
           end
-          options = @options.dup.merge(server_options)
-          options.delete(:digest_class)
-          Dalli::Server.new(s, options)
```
mdehoog, line 358: just a cleanup... the `dup` is not necessary as `merge` creates a new Hash... happy to re-add the `:digest_class` key deletion if necessary, but `Dalli::Server` doesn't use this key

Issue #741, opened 2020-03-09
text protocol and meta command support


shivnagarajan:
With the addition of the new [metacommands](https://github.com/memcached/memcached/wiki/MetaCommands) to memcached (available in the text protocol), would there be interest to support the text protocol with dalli? (and supporting the meta commands with it also).

Addition of text protocol support would also allow the use of dalli with proxies like [mcrouter](https://github.com/facebook/mcrouter).

Are there any caveats/issues to be concerned about in adding this support. 

We use dalli, (in some cases), text protocol  (in many cases) mcrouter and currently are working towards using the meta commands. As it stands our applications that use dalli cannot use mcrouter, and applications that use text protocol (and mcrouter) cannot use dalli, and these clients will soon be adding the use of the meta commands also.

@ndurks @dormando

dormando:
I wonder if it makes sense for this to be done with dalli, since that was always a binprot-first client. You might have better luck with a compatibility wrapper over something really minimalist/simple.

mperham:
Dalli is in long-term maint mode, we don't have a maintainer that can spend a lot of time on major overhauls. You can fork Dalli or you can step up if you have a vision for 3.0 and are willing to do the work.

byroot:
> are willing to do the work

Of course. I'll submit a few PRs soonish. Also I've seen https://github.com/petergoldstein/dalli/issues/758, I can help with that TODO list.

PR #760, opened 2020-09-27
Allow to pass an alternative implementation of Dalli::Server

casperisfine:
As discussed in https://github.com/petergoldstein/dalli/issues/741

This is all I need to maintain an external gem that make Dalli use the new "meta commands" protocol, whithout monkey patches.

If you are interested I can also submit a PR with said protocol support and pledge to maintain it, but that's up to you.

@mperham


mperham:
Want to rename Dalli::Server to something more applicable, like Dalli::Protocol::Binary?

casperisfine:
> Want to rename Dalli::Server to something more applicable, like Dalli::Protocol::Binary?

Yeah I had that in mind but was afraid it would be too big of a diff. I just updated the PR do to that.

Does this mean you'd be interested in including my `meta commands` protocol support? I can PR it once this is merged if you'd like.

casperisfine:
jruby CI hit a timeout, but other than that it's green.

mperham:
I’m ok with abstracting the protocol handler so you can override it but I’m not interested in maintaining a separate impl.

casperisfine:
Fine by me.

petergoldstein:
@casperisfine It's been a long time, but are you still interested in getting this in?  If so, would you mind rebasing?  We can then work through any issues.

byroot:
> @casperisfine It's been a long time, but are you still interested in getting this in?

Yes I am still interested. Unfortunately I'm away until Wednesday, but I'll happily rebase then.

casperisfine:
@petergoldstein  It should be good to go now.

petergoldstein:
@casperisfine Great.  Will take a look today or tomorrow.

Any thoughts on whether this change requires a minor or major release?  Or do you think we can put it in a patch?  @mperham  do you have any opinion?

casperisfine:
Hum, is `Dalli::Server` a public API? If not, it's an implementation detail.

Regardless we could perfectly alias `Dalli::Protocol::Binary` as `Dalli::Server` and even emit a deprecation warning, Rails does this routinely.

mperham:
That’s my only concern, renaming D::Server. Probably a fair amount of people monkeypatching that class.

petergoldstein:
I THINK it's an implementation detail, so a patch should be fine.  But there may be gems out there that monkeypatch.  But I haven't been able to find anything in a quick search.

casperisfine:
An alias is easy to add, and I think we should just to be safe.

The question is wether you want it to emit a deprecation warning or not. Just let me know and I'll add it.

petergoldstein:
@casperisfine Sounds good.  If you could add the alias with a deprecation warning, that would work.  Thanks.

casperisfine:
Done.

petergoldstein:
This looks good.  Merging.

petergoldstein:
Thanks much @casperisfine .  Sorry it took so long.  I should have a patch release out this week which includes this.

casperisfine:
No worries, thanks for the merge!

PR #762, opened 2020-09-28
Avoid minitest assert usage deprecation warning

olleolleolle:
The warning looked like: `DEPRECATED: global use of must_raise from /home/travis/build/petergoldstein/dalli/test/test_server.rb:93. Use _(obj).must_raise instead. This will fail in Minitest 6.`

### DIFF
--- a/test/test_server.rb
+++ b/test/test_server.rb
@@ -91,10 +91,10 @@ describe Dalli::Server do
     end
 
     it "throws an exception if the hostname cannot be parsed" do
-      lambda { Dalli::Server.new("[]") }.must_raise Dalli::DalliError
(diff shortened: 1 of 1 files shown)


olleolleolle:
The failing test is the JRuby one, about a timeout. That file has not been modified in this PR.

```
  1) Failure:
failover::timeouts#test_0001_not lead to corrupt sockets [/home/travis/build/petergoldstein/dalli/test/test_failover.rb:16]:
Did not timeout in 2.333481
```

mperham:
Sorry, I fixed this myself before realizing you had submitted a PR to do the same.

PR #764, opened 2020-10-21
Document `:mem_cache_store` `$MEMCACHE_SERVERS` incompatibility

sambostock:
This adds a warning to `History.md` and `3.0-Upgrade.md` to help prevent mistakes when migrating from `:dalli_store` to `:mem_cache_store`.

`:dalli_store` falls back to `$MEMCACHE_SERVERS` if no addresses are given, whereas `:mem_cache_store` goes straight to `localhost:11211`.

This is not immediately obvious, especially to users not providing any custom options, and makes it easy to accidentally effectively disable caching entirely (if `localhost:11211` is not a memcached server).

Note that rails/rails#40420 would add support for `$MEMCACHE_SERVERS`. Nonetheless, I think it would make sense to get this documentation out ASAP to prevent accidents. If the PR is accepted, it would make sense to follow up with a message mentioning specific Rails versions, once it has made it into a release.

### DIFF
--- a/3.0-Upgrade.md
+++ b/3.0-Upgrade.md
@@ -5,6 +5,8 @@ This major version update contains several backwards incompatible changes.
 * **:dalli_store** has been removed. Users should migrate to the
   official Rails **:mem_cache_store**, documented in the [caching
 guide](https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-memcachestore).
+  Note that `:mem_cache_store` does _not_ check for `ENV['MEMCACHE_SERVERS']`
+  prior to Rails 6.1, so you may need to explicitly provide it.
 * Attempting to store a larger value than allowed by memcached used to
   print a warning and truncate the value. This now raises an error to
   prevent silent data corruption.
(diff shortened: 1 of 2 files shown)


sambostock:
Well, that was fast. rails/rails#40420 has been merged already, and should be included as part of the Rails 6.1 release. I've updated the docs accordingly.

sambostock:
@petergoldstein what do you think of this change?

sambostock:
Maybe I should have pinged @mperham instead?

mperham:
It’s never something Dalli has documented and was a legacy compat thing
from the previous memcache gem.

sambostock:
Thanks for the context @mperham.

While the behaviour isn't documented in `:dalli_store`, it doesn't log deprecation warnings either, so (having little Memcached experience) it wasn't something I thought of as unusual.

My concern is that making this change can break things without warning

```diff
-config.cache_store = :dalli_store
+config.cache_store = :mem_cache_store
```

I was hoping to prevent others from encountering the same issue by calling it out in the migration guide.

PR #767, opened 2020-12-10
Ruby-prof takes an IO object for print nowadays

nateberkopec:
Code as written blows up on latest rubyprof, they changed print to take an IO instead


petergoldstein (approved):
Looks good.  Thanks

Issue #769, opened 2020-12-27
Support for gat command


tbeauvais:
The "gat" (get and touch) command was added in Memcached 1.4.8. This is pretty handy and allows us to avoid performing multiple commands.

petergoldstein:
Addressed with the merge and release of #770

PR #770, opened 2020-12-27
Add gat command support

tbeauvais:
Add support for "gat" (get and touch) fetch an item and simultaneously update its expiration time.
See #769


petergoldstein (approved):
Thanks for the contribution!

PR #777, opened 2021-07-26
Allow to initialize Client with just options and default server

semaperepelitsa:
If you try to specify options without a server, you get confusing errors - until you notice that your options were interpreted as servers (with hash converted to an array). 

```
irb(main):011:0> m = Dalli::Client.new(namespace: "test")
=> #<Dalli::Client:0x000000009a9c5820 @servers=[:namespace, "test"], @options={}, @ring=nil>
irb(main):012:0> m.get("foo")
Dalli::Server#connect namespace:11211
namespace:11211 failed (count: 0) SocketError: getaddrinfo: hostname nor servname provided, or not known
Dalli::Server#connect namespace:11211
namespace:11211 failed (count: 1) SocketError: getaddrinfo: hostname nor servname provided, or not known
namespace:11211 is down
Dalli::Server#connect test:11211
test:11211 failed (count: 0) SocketError: getaddrinfo: hostname nor servname provided, or not known
Dalli::Server#connect test:11211
test:11211 failed (count: 1) SocketError: getaddrinfo: hostname nor servname provided, or not known
test:11211 is down
down_retry_delay not reached for test:11211 (60.000 seconds left)
down_retry_delay not reached for namespace:11211 (59.989 seconds left)
down_retry_delay not reached for namespace:11211 (59.989 seconds left)
down_retry_delay not reached for namespace:11211 (59.988 seconds left)
down_retry_delay not reached for test:11211 (60.000 seconds left)
down_retry_delay not reached for namespace:11211 (59.988 seconds left)
down_retry_delay not reached for test:11211 (59.999 seconds left)
down_retry_delay not reached for namespace:11211 (59.988 seconds left)
down_retry_delay not reached for namespace:11211 (59.988 seconds left)
down_retry_delay not reached for namespace:11211 (59.988 seconds left)
down_retry_delay not reached for test:11211 (59.999 seconds left)
down_retry_delay not reached for test:11211 (59.999 seconds left)
down_retry_delay not reached for namespace:11211 (59.987 seconds left)
down_retry_delay not reached for namespace:11211 (59.987 seconds left)
down_retry_delay not reached for test:11211 (59.998 seconds left)
down_retry_delay not reached for namespace:11211 (59.987 seconds left)
Traceback (most recent call last):
Dalli::RingError (No server available)
```

You can work around that by passing nil as the first argument:`Dalli::Client.new(nil, namespace: "test")`. But this small patch will solve the issue altogether and make Client more intuitive to use.

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -31,9 +31,14 @@ module Dalli
     # - :cache_nils - defaults to false, if true Dalli will not treat cached nil values as 'not found' for #fetch operations.
     # - :digest_class - defaults to Digest::MD5, allows you to pass in an object that responds to the hexdigest method, useful for injecting a FIPS compliant hash object.
     #
-    def initialize(servers = nil, options = {})
+    def initialize(servers = nil, options = nil)
+      if options.nil? && servers.is_a?(Hash)
+        options = servers
+        servers = nil
+      end
+
       @servers = normalize_servers(servers || ENV["MEMCACHE_SERVERS"] || "127.0.0.1:11211")
-      @options = normalize_options(options)
+      @options = normalize_options(options || {})
       @ring = nil
     end
 
--- a/test/test_dalli.rb
+++ b/test/test_dalli.rb
@@ -104,6 +104,16 @@ describe "Dalli" do
     assert_equal s2, s3
   end
 
+  it "can set options with default server" do
+    dc = Dalli::Client.new(compress: true)
+    assert dc.instance_variable_get(:@options)[:compress]
+
+    ring = dc.send(:ring)
+    assert_equal 1, ring.servers.size
+    s1 = ring.servers.first.hostname
+    assert_equal "127.0.0.1", s1
+  end
+
   it "accept comma separated string" do
     dc = Dalli::Client.new("server1.example.com:11211,server2.example.com:11211")
     ring = dc.send(:ring)


petergoldstein:
@semaperepelitsa Honestly, this solution strikes me as a little too clever in interpreting the caller's intent and, as such, confusing in practice.

To resolve your very reasonable complaint about the confusing errors I'd suggest explicit validation of the type of the argument(s).  Specifically, `servers` should be an array if it isn't nil, and should raise an `ArgumentError` with a message like "The servers argument should be nil or an array of servers".

petergoldstein:
@semaperepelitsa I put up a PR - #786  - that implements the ArgumentError approach.  Let me know if that would resolve your concern about confusing behavior.

semaperepelitsa:
That works for avoiding the mistake, but still forces you to pass the first argument even if you don't want to change the default servers. Ideally, it should become its own option `:servers`, so the entire issue could be avoided. 

```
Dalli::Client.new(nil, namespace: "test")
```

It is still an improvement over the current situation, so I support it.

petergoldstein:
@semaperepelitsa I get that, but I don't think changing the signature of initialize is in the cards in the near future.  Thanks for raising the issue.

petergoldstein:
Closing in favor of #786

semaperepelitsa:
I understand. Thanks for looking into this!

PR #781, opened 2021-08-31
Add Ruby 3.0 to test matrix and see what breaks

petergoldstein:


petergoldstein:
Apparently nothing breaks.  So let's add this to the test matrix, and we can cut a release with official support for Ruby 3.

PR #754, opened 2020-09-04
Cathal/get multi should retry

ccoffey:
`cache.read` automatically retries on `NetworkError` but `cache.read_multi` does not which is unexpected.

The below results were generated by running the server as follows: 
`memcached -o idle_timeout=2`

The following setup was also run:
```ruby
test_memcache_store = 'localhost:11211'

test_memcache_settings = {
  expires_in: 3.days,
  failover: false,
  socket_timeout: 5,
  down_retry_delay: 0,
  socket_max_failures: 5,
  compress: true,
  pool_size: 30,
  pool_timeout: 5,
  value_max_bytes: 104857600,
  instrument_errors: true,
}

test_store = :dalli_cas_store, test_memcache_store, test_memcache_settings

cache = ActiveSupport::Cache.lookup_store(test_store)

# Fill the cache with 10 values
(1..10).each { |i| cache.write(i.to_s, "value#{i}") }
```

Before this PR:

```ruby
# Wait for 5 seconds so that the open connection becomes idle and is closed by the server
cache.read("1")
Cache read: 1
localhost:11211 failed (count: 0) EOFError: end of file reached
#<Dalli::NetworkError: Socket operation failed, retrying...>
retrying request with new server
Dalli::Server#connect localhost:11211
=> "value1"
```

```ruby
# Wait for 5 seconds so that the open connection becomes idle and is closed by the server
cache.read_multi("1")
Cache read_multi: ["1"]
localhost:11211 failed (count: 0) EOFError: end of stream
=> {}
```

**Note:** `read_multi` does not retry and returns `{}`

After this PR:

```ruby
# Wait for 5 seconds so that the open connection becomes idle and is closed by the server
cache.read("1")
Cache read: 1
localhost:11211 failed (count: 0) EOFError: end of file reached
#<Dalli::NetworkError: Socket operation failed, retrying...>
retrying request with new server
Dalli::Server#connect localhost:11211
=> "value1"
```

```ruby
# Wait for 5 seconds so that the open connection becomes idle and is closed by the server
cache.read_multi("1")
Cache read_multi: ["1"]
localhost:11211 failed (count: 0) EOFError: end of stream
#<Dalli::NetworkError: Socket operation failed, retrying...>
retrying request with new server
Dalli::Server#connect localhost:11211
=> {"1"=>"value1"}
```

Issue #390, opened 2013-09-17
Dalli returning true:TrueClass on memcached node reboot


mheffner:
Seems related to #146, but we appear to see a similar problem with the latest dalli when a memcached node is rebooted.

When we reboot a node in our Elasticache, memcached 1.4.5, cluster under load we see numerous failures from `get` returning `true:TrueClass` instead of the cached object (or simply nil).

memcached 1.4.5
dalli 2.6.4
ruby 1.9.3-p429

Using the connection settings:

```ruby
    CLIENT_SETTINGS = {
      # Default TTL
      :expires_in => 3600,
      # Don't move keys to failed hosts
      :failover => false,
      :down_retry_delay => 30
    }
```

Unfortunately I haven't been able to come up with a case that reproduces this. Anything to check to help debug this?

petergoldstein:
This issue never got much follow up, and there have been no other similar reports.  As the issue hasn't been commented on in over 2 years, I'm going to close.

@mheffner Please reopen if this is still an issue.  Thanks!

cryptomail:
We can confirm we were afflicted with this.  There appears to be a state on AWS Elasticache that is 1/2 baked, where the server accepts connections, yet has no data initialized and then simply returns status === 0 with no data 🥰 
It's still an issue and my partner @snbloch and I tried very naively to fix it but we blew all the tests up lol.
Honestly we feel it's a problem with memcached implementors.  Why accept connections and afford ANY response when you (sorry for the anthropomorphism) haven't loaded all your data yet, and truly aren't ready?!
I'm still wondering if there is some path that can detect this state in Dalli, and afford an error state to clients.

Our use case happens to be very fortunate in that we can detect this state higher up as a client of this library, and can rule out the silly True Object from being returned, but that's no way to live really either :)

Thank you for your continued support on Dalli.

PR #783, opened 2021-09-30
Raise NetworkError and reconnect when multi_response_nonblock is unexpectedly called

mervync:
When an elasticache node is rebooted, it seems to send invalid response headers that cause dalli to think that the multiget response has completed, when there is actually still data queued up in the socket.

This causes dalli to raise `RuntimeError, "multi_response has completed"`, which is acceptable behavior. However, the problem is that the data queued up in the socket then **leaks into subsequent responses**.

It is unclear if this is an elasticache-specific bug. This may be related to the issue reported in #390.

To work around this issue, raise NetworkError and reconnect whenever we run into this state. The multiget response will be incomplete but it seems like this is being addressed in #754.

Other possible solutions:

- Call `down!` - this also works, but may cause the server to be marked as down for a longer time than necessary.
- Attempt to drain the socket when this happens - seems like a rather more complicated fix.

### DIFF
--- a/lib/dalli/server.rb
+++ b/lib/dalli/server.rb
@@ -145,7 +145,7 @@ module Dalli
     #
     # Returns a Hash of kv pairs received.
     def multi_response_nonblock
-      raise "multi_response has completed" if @multi_buffer.nil?
+      reconnect! "multi_response has completed" if @multi_buffer.nil?
 
       @multi_buffer << @sock.read_available
       buf = @multi_buffer
--- a/test/test_server.rb
+++ b/test/test_server.rb
@@ -191,4 +191,24 @@ describe Dalli::Server do
       end
     end
   end
+
+  describe "multi_response_nonblock" do
+    subject { Dalli::Server.new("127.0.0.1") }
+
+    it "raises NetworkError when called before multi_response_start" do
+      assert_raises Dalli::NetworkError do
+        subject.request(:send_multiget, ["a", "b"])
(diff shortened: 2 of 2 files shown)


petergoldstein:
@mervync in your opinion would we need to ensure that #754 is merged along with this submission for correct behavior?  Or are the two independent?

And are there other known situations where we're likely to run into this error?

mervync:
Other networking-related issues could cause the issue described in #754 to arise, so the two issues seem independent to me. 

I didn't notice any other cases where this error could arise. That said, the code around multi gets is quite complex...

petergoldstein:
@mervync This looks right to me.

The issue seems to potentially arise when either i) There's a client side abort or ii) The server sends a bad key/value header.  In either case the natural thing to do is reconnect the socket, since the state of data on the socket is corrupt or subject to a possible race condition.  "Normal" operations should be unaffected by this change.

As your tests note, there's also the case where multi_response_start hasn't been called beforehand.  One could argue this case should be handled with an error (as the socket isn't even connected), but given the usage of that method I'm satisfied with the proposed approach.

Thanks for the contribution.

mervync:
@petergoldstein Thanks for reviewing and merging. Would greatly appreciate a patch release 🙏

petergoldstein:
@mervync I need to review changes since last patch, and I want to see if we can get one more PR in, but I should do a patch release reasonably soon.

PR #788, opened 2021-10-13
read_multi should clean connections and retry on NetworkError

andrejbl:
This PR supersedes the change proposed in https://github.com/petergoldstein/dalli/pull/754 (@ccoffey and myself work in the same team at Intercom). It is trying to tackle the same issue, namely the lack of retries in `read_multi` when a pool of memcached servers have `idle_timeout` set. Opening a new PR as I don't have access to the old fork from which the original change was proposed.

Copying the repro here for clarity:

The below results were generated by running the server as follows:
```
memcached -o idle_timeout=2
```

The following setup was also run:

```ruby
test_memcache_store = 'localhost:11211'

test_memcache_settings = {
  expires_in: 3.days,
  failover: false,
  socket_timeout: 5,
  down_retry_delay: 0,
  socket_max_failures: 5,
  compress: true,
  pool_size: 30,
  pool_timeout: 5,
  value_max_bytes: 104857600,
  instrument_errors: true,
}

test_store = :dalli_cas_store, test_memcache_store, test_memcache_settings

cache = ActiveSupport::Cache.lookup_store(test_store)

# Fill the cache with 10 values
(1..10).each { |i| cache.write(i.to_s, "value#{i}") }
```

Before this PR:

```ruby
# Wait for 5 seconds so that the open connection becomes idle and is closed by the server
cache.read("1")
Cache read: 1
localhost:11211 failed (count: 0) EOFError: end of file reached
#<Dalli::NetworkError: Socket operation failed, retrying...>
retrying request with new server
Dalli::Server#connect localhost:11211
=> "value1"
# Wait for 5 seconds so that the open connection becomes idle and is closed by the server
cache.read_multi("1")
Cache read_multi: ["1"]
localhost:11211 failed (count: 0) EOFError: end of stream
=> {}
Note: read_multi does not retry and returns {}
```

After this PR:

```ruby
# Wait for 5 seconds so that the open connection becomes idle and is closed by the server
cache.read("1")
Cache read: 1
localhost:11211 failed (count: 0) EOFError: end of file reached
#<Dalli::NetworkError: Socket operation failed, retrying...>
retrying request with new server
Dalli::Server#connect localhost:11211
=> "value1"
# Wait for 5 seconds so that the open connection becomes idle and is closed by the server
cache.read_multi("1")
Cache read_multi: ["1"]
localhost:11211 failed (count: 0) EOFError: end of stream
#<Dalli::NetworkError: Socket operation failed, retrying...>
retrying request with new server
Dalli::Server#connect localhost:11211
=> {"1"=>"value1"}
```

We have seen that the approach with using simple `raise` when connections are opened to multiple servers and one connections idles out hits a problem because `@inprogress` and other response state variables on the protocol side are left untouched and cause exceptions when the code tries to resend the requests. The solution proposed here is to abort all of the connections when retrying on `NetworkError` and let the `perform` re-establish them again. It is a simpler logic, at the expense of some connection churn.

The PR also fixes what seems to be a bug in the `perform_multi_response_start` when there are multiple servers - deleting a server within the same block that iterates over the server collection causes iterator to incorrectly iterate. The deletion is now done outside of the loop.

### DIFF
--- a/test/test_failover.rb
+++ b/test/test_failover.rb
@@ -49,14 +49,15 @@ describe "failover" do
         memcached(port_2, "-o idle_timeout=1") do |_, second_port|
           dc = Dalli::Client.new ["localhost:#{first_port}", "localhost:#{second_port}"]
           dc.set "foo", "bar"
-          foo = dc.get "foo"
-          assert_equal foo, "bar"
+          dc.set "foo2", "bar2"
+          foo = dc.get_multi "foo", "foo2"
+          assert_equal foo, { "foo"=>"bar", "foo2"=>"bar2" }
 
           # wait for socket to expire and get cleaned up
-          sleep 1.2
+          sleep 5
 
-          foo = dc.get "foo"
-          assert_equal foo, "bar"
+          foo = dc.get_multi "foo", "foo2"
+          assert_equal foo, { "foo"=>"bar", "foo2"=>"bar2" }
         end
       end
     end


test/test_binary_protocol.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "ostruct"
```
petergoldstein, line 3: @andrejbl As no tests were changed/added, why is this line required?

petergoldstein:
@andrejbl So my initial take is that the changes look good and make sense.  Are there any tests we can add to ensure things behave as expected?

PR #790, opened 2021-10-13
Fix SyntaxError: unterminated string meets end of file

ryanfb:
This is meant to be a deprecation warning, but currently any code that loads this file will raise a SyntaxError.


petergoldstein:
@ryanfb Thanks for the fix.  I'll release a 3.0.1 with this fix today.

PR #795, opened 2021-10-14
Start adding Rubocop

petergoldstein:
This is a spike on adding Rubocop to Dalli, using the default Rubocop configuration as much as possible.  The goals are:

1. To ensure that Dalli's codebase matches current best practices for writing idiomatic Ruby.
2. To expose issues with the code or tests via linting, now and ongoing
3. To make the code base more approachable for the general developer population

As part of this effort, the following issues have been identified to date:

1. Two specs that, because they were using `assert` rather than `assert_equals` were not actually functional.
2. The `find_memcached` code only checked the $PATH, rather than checking a set of additional standard paths, as was intended.
3. Existing use of IO.select that's incompatible with the fiber scheduler in Ruby 3.0 and forward

The bulk of the changes are expected to be cosmetic, and reflect the changing nature of idiomatic Ruby over time.  I'm going to try and avoid other refactoring - designed to shrink the size and complexity of classes/methods/blocks - within this PR. 
 That will likely require that we use non-default Rubocop configuration in some places.


dalli.gemspec
```diff
@@ -15,6 +17,12 @@ Gem::Specification.new do |s|
...
   s.homepage = "https://github.com/petergoldstein/dalli"
-  s.add_development_dependency "rack"
+  s.required_ruby_version = '>= 2.5'
+
   s.add_development_dependency "connection_pool"
+  s.add_development_dependency "rack"
+  s.add_development_dependency "rubocop"
+  s.add_development_dependency "rubocop-minitest"
+  s.add_development_dependency "rubocop-performance"
+  s.add_development_dependency "rubocop-rake"
```
olleolleolle, line 27: If all development is happening with Bundler, perhaps no development deps are necessary to add in the gemspec?

petergoldstein: Maybe not.  I don't have a strong opinion on that one way or the other.  Not sure it hurts.

petergoldstein:
The current PR doesn't fix all lints, but most of the remaining issues are in areas that I expect to see some refactoring in other PRs.  So I'm inclined to land this PR on master at this point, and use that as the base going forward.

PR #798, opened 2021-10-15
Compression updates - enable/disable per request + documentation

petergoldstein:
Update compression calculation to enable per-request enable/disable of compression.

Update compression related documentation to be correct.


petergoldstein:
@rafaelfranca I believe this change will address the concern you expressed here - https://github.com/petergoldstein/dalli/commit/e9afa45c5f8a134c4a7ca853530eb0ff8aeb3e01#r58017079

Can you please comment?  Thanks.

petergoldstein:
@rafaelfranca I did a little refactoring - extracted a class - and added comprehensive tests for all option combinations.  I'm not nuts about the class name, but I haven't come up with anything better that wouldn't be confusing.  So I'm going to sit on this for a day or so and then merge, at which point I'll release.  Please let me know if you have any comments.

PR #821, opened 2021-11-02
Prepare to Extract MultiKeyProcessor

petergoldstein:
Much of the complexity of the Dalli::Client class is in the `get_multi_yielder` code. 

The client implementation is overwhelmingly focused on the no-block version perform method - specifically taking client level calls, resolving them to the correct server, and invoking the appropriate protocol operation.   There's a small amount of additional logic for parsing arguments and initializing the Ring.

Essentially a multi-yield operation is a very different operation, and there's a lot of logic to ensure it handles errors correctly, produces expected results, etc.  Pulling that logic into a separate class, invoked by the client, will make it easier to understand and modify.

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -370,52 +370,6 @@ module Dalli
       raise ArgumentError, "Cannot convert ttl (#{ttl}) to an integer"
     end
 
-    def groups_for_keys(*keys)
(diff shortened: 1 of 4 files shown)


petergoldstein:
I'm getting a repeatable failure in CI, but locally everything runs fine.  Will need a look.

PR #827, opened 2021-11-05
Reenable SASL tests on CI

petergoldstein:


petergoldstein:
Replaced with an alternate implementation

PR #828, opened 2021-11-12
Switch to inheriting from Rack's Abstract::PersistedSecure

### LINKED ISSUE #765: `session[:session_id]` return undefined method `public_id' with Rails 5.2
I'm upgrading an application from Rails 5.1 to Rails 5.2. When I do `session[:session_id]`, it gives the error `undefined method 'public_id'` because of a change introduced in [this commit](https://github.com/rails/rails/commit/92ec9f270d696a17ba7604b3d02931f07c3e7c6a). `session[:session_id]` is doing `session.id.public_id` [internally](https://github.com/rails/rails/blame/5699122abf7613465c62f2e50e3af017e95cb609/actionpack/lib/action_dispatch/request/session.rb#L95-L99). It works correctly with `ActionDispatch::Session::CookieStore` because, in this case, `session.id` returns a [Rack::Session::SessionId](https://github.com/rack/rack/blob/2.1.4/lib/rack/session/abstract/id.rb) which implement `public_id`.

`session.id` should probably return and object that implement the same methods as `Rack::Session::SessionId`.

petergoldstein:
This update changes the Dalli session store implementation to inherit from Abstract::PersistedSecure rather than Abstract::Persisted.  Effectively this change decouples the public id value (the one placed in the cookie) from the key used for internal lookups in memcached.

This should fix #765 .

It is also likely a breaking change, as it will invalidate any outstanding sessions.


petergoldstein:
@GCorbel I know it's been a while since you reported #765 , but if you're still around and so inclined it would be helpful to confirm that this fix addresses your issue.  Thanks.

GCorbel:
Hi, I give it a try on a Rails 6.0.4.1 application. I just changed to use : 

```
gem 'dalli', github: 'petergoldstein/dalli', branch: 'feature/switch_to_persisted_secure'
```

And added `session[:session_id]` in a controller and I still have `NoMethodError: undefined method `public_id' for "13ef84661f6d4cf846edfe8ac5a7fd03":String`. I checked, and the application really use your code.

petergoldstein:
Thanks @GCorbel .  Any chance you'd be able to post your test app in a repo?  That would make it easier to nail this down.

petergoldstein:
@GCorbel You can ignore my last request.  I believe I found the issue and so I don't need a test app.  If you could rerun your test with the latest code from this PR, I'd appreciate it.  Thanks.

GCorbel:
I confirm it works. `session.id.class` gives `Rack::Session::SessionId` and `session[:session_id]` gives the correct id.

petergoldstein:
@GCorbel Great.  I'll merge and release (probably as 3.1.0) soon.

lib/rack/session/dalli.rb
```diff
@@ -78,95 +66,116 @@ def initialize(app, options = {})
...
+
+      def build_data_source(options)
+        server_configurations, client_options, pool_options = extract_dalli_options(options)
+
+        if pool_options.empty?
+          ::Dalli::Client.new(server_configurations, client_options)
+        else
+          ensure_connection_pool_added!
+          ConnectionPool.new(pool_options) do
+            ::Dalli::Client.new(server_configurations, client_options.merge(threadsafe: false))
```
bf4, line 134: I can't tell why the args here to changed to be always `threadsafe: false` in a connection pool

to

```ruby          ConnectionPool.new(pool_options) do
            ::Dalli::Client.new(server_configurations, client_options.merge(threadsafe: false))
```

from

```ruby
        mserv, mopts, popts = extract_dalli_options(options)
# where essentially has
#        if mopts[:pool_size] || mopts[:pool_timeout]
#          mopts[:threadsafe] = true
#        end

        @pool = ConnectionPool.new(popts || {}) { ::Dalli::Client.new(mserv, mopts) }
```

context: we have this setup and seeing (sometimes, rarely, in spurts) cache reads get the wrong data

petergoldstein: @bf4 The whole notion of using a connection pool is that individual threads check out a connection and then check it back in when they are done with it.  While it is checked out it is unavailable for use by any other thread.  Ergo, any thread safety at the `Dalli::Client` is superfluous.

This is a relatively standard usage and is unlikely to be the source of your data corruption issues.  If you can produce a reproducible case I'll take a look.

bf4: Thanks @petergoldstein for the quick response.   I figure at least now the reason for the change is sufficiently commented :)

I don't expect to have a reproducible case at this point.  I've done quite a bit of logging and this was my shot in the dark.  Guess it's either gremlins or some unreported bug at redislabs or some network shenanigans at Heroku .

PR #836, opened 2021-11-17
Extract response processing into a separate class

petergoldstein:
- Reorganize binary.rb to separate response code
- Extract some minor response parsing code from incr/decr and pull into a method
- Put NOT_FOUND in top level Dalli module, preserve reference for backwards compatibility
- Rename and encapsulate @inprogress


petergoldstein:
I'm happy with this extraction, even if there's some future class reorganization required, so I'm going to merge.  Simplifying the response processing logic and pulling it out into a dedicated class feels like a win.

Issue #865, opened 2021-12-13
Renaming of multi? to quiet? can be breaking


larouxn:
👋 Just a heads up issue that the recent [3.1.1 patch version](https://github.com/petergoldstein/dalli/blob/b5ab6cda291edd98476b0f9dc04ed444d1e6c61b/History.md#311) release contained a rename of `multi?` to `quiet?` which was a breaking change (no method error) for a service I was running. Not sure if that will be the case for many apps/services but just wanted to give a heads up. Maybe we can add a note to the `History.md` warning people of potential breakage.

petergoldstein:
Let me check on that.  I thought I'd aliased all the public methods

petergoldstein:
@larouxn  So `multi?` wasn't a method on `Dalli::Client`.  How is your service utilizing this method?  Are you directly accessing the `Dalli::Protocol::Binary`?

larouxn:
@petergoldstein, the calls are indeed directly on `Dalli::Protocol::Binary`. Still trying to track down exactly what is making the `multi?` calls to see if it can either stop making calls on `Binary` or address the outdated method name.

```
NoMethodError · undefined method `multi?' for #<Dalli::Protocol::Binary:0x00007fd81ed3e4d0>
```

If it's a public gem making the `multi?` calls, I hope we can address the issue. Will report back here if I find anything.

larouxn:
👋 Alright, I was able to find the `multi?` method usage. Thankfully just an internal library, not in a public gem so this may be a false alarm re: breaking change. Granted, other people outside my organization may also be using that method directly, in which case would still be breaking for them. Maybe still worth noting.

petergoldstein:
@larouxn Ok, the `multi?` method on `Dalli::Protocol::Binary` was private.  It's definitely not something that users of the library should be calling in general.  Can you share why your internal library was calling it?

larouxn:
It's used in an instrumentation library that I'm looking at for the first time right now. Seems to be used to ensure we don't instrument Dalli call latency during multi calls such as `multi_get` as that would be statistically inaccurate compare to non-multi methods calls. Seems normal `multi` is also used for "quiet operations".

Hoping to tag someone with more context into here.

robertlaurin:
[...]

PR #866, opened 2021-12-13
Make quiet? and multi? public on the Dalli::Protocol::Binary class

petergoldstein:


robertlaurin (approved):
Thanks @petergoldstein, that should close https://github.com/petergoldstein/dalli/issues/865 for us.

PR #872, opened 2021-12-15
Minor response parsing improvements

casperisfine:
- Get rid of an useless String#byteslice. `unpack1('N')` doesn't care if there's more bytes left.
- Use `String#b` to directly create a binary buffer. Saves some more encoding checks.
- Favor `byteslice` over `slice` and `#[]` as it doesn't need to check encoding as much, and allows shared strings.

For context I wanted to use [the new `unpack(offset: )` to save more allocations](https://bugs.ruby-lang.org/issues/18254), but it's not as easy with the recent refactor. I can look more into it if there's interest.

### DIFF
--- a/lib/dalli/protocol/binary/response_processor.rb
+++ b/lib/dalli/protocol/binary/response_processor.rb
@@ -47,9 +47,9 @@ module Dalli
         end
 
         def unpack_response_body(extra_len, key_len, body, unpack)
-          bitflags = extra_len.positive? ? body.byteslice(0, extra_len).unpack1('N') : 0x0
+          bitflags = extra_len.positive? ? body.unpack1('N') : 0x0
           key = body.byteslice(extra_len, key_len) if key_len.positive?
-          value = body.byteslice(extra_len + key_len, body.bytesize - (extra_len + key_len))
+          value = body.byteslice((extra_len + key_len)..-1)
           value = unpack ? @value_marshaller.retrieve(value, bitflags) : value
           [key, value]
         end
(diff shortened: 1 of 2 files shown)


petergoldstein:
Thanks @casperisfine .

For what it's worth, I'm not sure the `unpack(offset: ...)` pattern would be usable in Dalli in any case, given that we need to support Ruby 2.5+ and JRuby.

casperisfine:
> I'm not sure the `unpack(offset: ...)` pattern would be usable in Dalli in any case, given that we need to support Ruby 2.5+ and JRuby.

The usual pattern is to test for the existence of the feature and then define two versions of the same method:

```ruby
if String.instance_method(:unpack).parameters.include?([:key, :offset])
  def blah
    # Optimized version
  end
else
  def blah
    # Fallback version
  end
end
```

That's how we get to use modern optimizations in Rails and many other places without having to drop support for old versions.

petergoldstein:
Sure, I get that.  But in my view that kind of code complexity tradeoff for what looks like a single allocation per packet is almost certainly not worth it.  In typical real world scenarios (where memcached is not collocated on the box with the client, where CPU/memory are not your gating factors, etc.), the performance gain from that change is not material (Amdahl's Law).  But the comprehensibility impact is significant.

At this point my focus is on simplifying and decoupling the code base.  That's exposed a number of long-standing bugs as well as some significant functional gaps.  Resolving the exposed bugs and addressing those functional gaps is in my view the highest value for effort in Dalli at this point.

casperisfine:
> But the comprehensibility impact is significant.

Yes, that's why I didn't do it, it was a very straightforward change before your latest refactor, it no longer is.

PR #879, opened 2021-12-17
Introduce the meta protocol as an option for Dalli

petergoldstein:
This PR introduces the meta protocol as an option for Dalli.  Configuring use of the meta protocol is done via the :protocol option in the client options.  Setting this to :meta will run with meta protocol.

As of this writing the meta protocol implementation should be complete.  All capabilities supported via binary are avavailable via meta with the exception of 1) SASL auth, which is not supported on meta and 2) delete_cas (see below)

The delete_cas functionality is not working because of an underlying memcached bug.  It is expected that this bug will be resolved in memcached shortly.


dormando:
fwiw re: SASL, there's a token based authentication in the text protocol. More similar to redis's basic one.

petergoldstein:
@dormando That's the auth supported by the '-Y' flag on startup?  Do people use it?

dormando:
Yes. I haven't a foggy. I just make it available :) SASL was always authentication and never encryption or serious security. Optional auth token usage can't hurt, and since it won't have platform weirdness like SASL adoption's probably easier.

petergoldstein:
I'm generally happy with this PR, and am going to publish it as 3.2.  This exercise was helpful in uncovering some significant testing gaps as well as a few existing bugs, fixes for which were published in 3.1.x versions.  While there are a couple of open issues in the repo, they aren't "bugs" per se - they are open questions/gaps that need answering.  So both the latest 3.1.x version and this version (which will be published as 3.2.0) should be fairly solid.

I haven't added token-auth, because I'm not sure how many systems/people actually use it.  If there's demand, it should be pretty straightforward to add.

PR #890, opened 2022-01-05
Changes to EOL Ruby 2.5 support

petergoldstein:


petergoldstein:
This should have been in 3.2.0 but I forgot.  Going to fold it into 3.2.1.

PR #901, opened 2022-02-11
Fix: null char not replaced

veritas1:
Aliyun returns a `null` char in the auth response which should be subbed out, otherwise PLAIN auth fails.


petergoldstein:
I'm going to merge and release this, even though I don't have a spec yet, because it's a fix for an obvious regression.

PR #917, opened 2022-06-15
Support old sessions

kbrock:
Hi Peter and team,

We have upgraded the version of `dalli` and the old sessions sometimes cause hiccups. Since we have a product installed on customer's sites, we are not totally in control if the customer remembers to flush the `memcached` daemon.

Is it possible to put something in place to more gracefully handle the case where the `dalli` gem has been updated but `memcached` has not been purged?

I wanted to put this out there to see if you were amiable to putting a fix into the core gem.

Thank you,

our current issue: https://github.com/ManageIQ/manageiq/issues/21772

### DIFF
--- a/lib/rack/session/dalli.rb
+++ b/lib/rack/session/dalli.rb
@@ -90,7 +90,7 @@ module Rack
 
       def delete_session(_req, sid, options)
         with_dalli_client do |dc|
-          dc.delete(memcached_key_from_sid(sid))
+          dc.delete(sid.respond_to?(:private_id) ? memcached_key_from_sid(sid) : sid)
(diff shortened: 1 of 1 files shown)


petergoldstein:
@kbrock I'm amenable to putting something like this in the gem and cutting a patch.  I'd like to make the change to the code a bit more resilient though - callers to memcached_key_from_sid should check for a nil value and not call the `Dalli::Client` in that case.  Let me put something up.

petergoldstein:
I merged in #918 in preference to this PR.  #918 has been released as part of 3.2.2

kbrock:
Thank you peter

PR #918, opened 2022-06-15
Avoid errors on old Dalli sessions

petergoldstein:
Right now older Dalli sessions references are causing code level errors because these sessions don't respond to private_id.  This change explicitly checks whether the sid responds to private_id before calling.  And it makes the callers handle a nil return value (when the sid doesn't respond to private_id)

This is an alternate to #917

### DIFF
--- a/lib/rack/session/dalli.rb
+++ b/lib/rack/session/dalli.rb
@@ -82,6 +82,9 @@ module Rack
       def write_session(_req, sid, session, options)
         return false unless sid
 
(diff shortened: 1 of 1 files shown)


petergoldstein:
@kbrock @Fryguy Let me know if you think this addresses your issue.

petergoldstein:
Consistent build failure is now sussed and resolved.  If this winds up green I'll merge.

PR #923, opened 2022-06-22
Evaluate Proc namespaces every time (not just at initialization)

nrw505:
When creating a dalli client with a Proc for the namespace option, it
would generally mean that the namespace may change from operation to
operation. Saving the result of calling the proc in the KeyManager
instance at initialisation time means that the client will always use
the value of the namespace as it was when the client was created. If
this was the desired behaviour, it would be just as easy to pass that
value as a string instead of a Proc.

Given test/test_key_manager.rb:69 is passing a namespace_as_symbol instead of namespace_as_proc, I'm assuming this is an unintended regression from when this logic was pulled out into KeyManager

### DIFF
--- a/lib/dalli/key_manager.rb
+++ b/lib/dalli/key_manager.rb
@@ -61,7 +61,7 @@ module Dalli
     def key_with_namespace(key)
       return key if namespace.nil?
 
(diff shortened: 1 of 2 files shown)


petergoldstein:
@nrw505 This all looks correct.  Thanks for catching the issue and contributing a fix.  Merging.

PR #925, opened 2022-07-19
Require "uri" from stdlib

adam12:
`URI` is used for parsing server config but is never explicitly required.


petergoldstein:
Good catch @adam12 .  Thanks for the contribution.

PR #928, opened 2022-08-04
Remove deprecation warning in client.rb

langsharpe:
Hi Team,

I was wondering if you would consider removing this warning from dalli? We use dalli as a dependency of a dependency. The middle dependency is unlikely to be updated (The PR is here https://github.com/nickelser/suo/pull/21). Would you consider removing this warning? As far as I can tell, there is no harm in leaving this file empty.

I'm not asking you to be forever backwards compatible. Just to remove this one grain of sand from our development environment.

Whatever you decide, thank you for maintaining this gem!

### DIFF
--- a/lib/dalli/cas/client.rb
+++ b/lib/dalli/cas/client.rb
@@ -1,3 +1,3 @@
 # frozen_string_literal: true
 
-puts "You can remove `require 'dalli/cas/client'` as this code has been rolled into the standard 'dalli/client'."
+# Empty file for backwards compatibility


petergoldstein:
Hi @langsharpe .  I appreciate that you'd like to see that deprecation message removed from your system, and that the `suo` gem isn't being updated so it's not likely to be fixed properly.  But this is a place where a deprecation warning makes a lot of sense.

There is actually some harm in leaving empty files around for historical reasons. It's one more thing that a contributor or maintainer needs to understand about the project history in order to understand the code's current state.  So I do want to make sure we clean up those things over time.  This deprecation warning let's users of the `Dalli` gem know that, at some point, this file will disappear (likely with the next major version bump).

So I'm going to close this PR, leaving the deprecation warning in place.  Thanks for the contribution, and I hope you understand my reasoning.

Haegin:
@petergoldstein would you be open to a PR changing `client.rb` to define a class with a method that emits the warning that it then calls? This would make it possible to monkey patch the deprecation warning out while keeping it around by default for users who would benefit from the warning. If you're open to that change let me know and I'll be happy to put a PR together.

petergoldstein:
@Haegin Yes.

PR #933, opened 2022-10-25
Fixes #932

### LINKED ISSUE #932: Meta protocol flush_all method is vulnerable to code injection (Lack of input type check)
Hi there,
I'm a security researcher currently doing research on Memcached wrappers vulnerabilities. I was doing some source code reviewing on your wrapper and noticed that the `flush_all` method on the meta protocol takes a delay value and passes it to the server without any checks, which can be used to smuggle commands to the Memcached server if an attacker has control over the value passed to the flush_all method.

https://github.com/petergoldstein/dalli/blob/5588d98f79eb04a9abcaeeff3263e08f93468b30/lib/dalli/protocol/meta.rb#L137-L140

https://github.com/petergoldstein/dalli/blob/5588d98f79eb04a9abcaeeff3263e08f93468b30/lib/dalli/protocol/meta/request_formatter.rb#L76-L81

### Proof of Concept
```ruby
require 'dalli'

# Proof of Concept by @xhzeem
$mcmeta = Dalli::Client.new('localhost:11211', protocol: :meta)
$mcmeta.set('xhzeem','meta')
$mcmeta.get("xhzeem") # b64_if_reg_match(\s)
puts $mcmeta.flush_all("\nset xhzeem 1 1000 8\ninjected") # :D
```

### Suggested Fix:
You should just add a simple check for the `delay` type and confirm it's a number or keep the 0 value.

petergoldstein:
Do some input sanitization for the meta protocol.  Since this protocol uses raw test, it is possible to execute an injection attack against unsanitized inputs.

This PR explicitly sanitizes the CAS arguments and the ttl passed to flush.

I also added some missing tests for `meta_arithmetic` and fixed a few lints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Dalli Changelog
 Unreleased
 ==========
 
+- Sanitize input to flush command to ensure additional commands are not passed to memcached (xhzeem / petergoldstein)
 - Namespaces passed as procs are now evaluated every time, as opposed to just on initialization (nrw505)
 - Fix missing require of uri in ServerConfigParser (adam12)
 - Fix link to the CHANGELOG.md file in README.md (rud)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Dalli Changelog
 Unreleased
 ==========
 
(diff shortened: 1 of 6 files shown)


petergoldstein:
@xhzeem I believe this will address the issue you raised.  Feel free to comment.  Thanks again.

PR #935, opened 2022-10-31
Changes for new empty line before assert cop.

petergoldstein:
This includes several spacing changes to satisfy the latest Rubocop::Minitest.

There is also a [bug in the latest version](https://github.com/rubocop/rubocop-minitest/issues/187) that is causing errors in the Rubocop run.  So my expectation is that this doesn't run green, but will be fixed once that bug is resolved.

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -30,6 +30,7 @@ module MiniTest
 
     def assert_error(error, regexp = nil, &block)
       ex = assert_raises(error, &block)
+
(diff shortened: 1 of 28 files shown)


petergoldstein:
I'm merging this, even though it's not green, because it addresses a number of lints.  I'm going to wait a bit to see if the underlying issue is fixed in `rubocop-minitest`. If not, I may simply refactor away the line causing the issue.

PR #936, opened 2022-11-24
Preallocate the read buffer in Socket#readfull

casperisfine:
Since we know exactly how many bytes we expect, instead of stating with an empty string and resizing many times, we can directly ask Ruby to preallocate the necessary space.

For large responses this saves many reallocations and copies.

### DIFF
--- a/lib/dalli/socket.rb
+++ b/lib/dalli/socket.rb
@@ -13,7 +13,7 @@ module Dalli
     ##
     module InstanceMethods
       def readfull(count)
-        value = +''
+        value = String.new(capacity: count + 1)
         loop do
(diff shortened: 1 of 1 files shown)


lib/dalli/socket.rb
```diff
@@ -13,7 +13,7 @@ module Socket
     ##
     module InstanceMethods
       def readfull(count)
-        value = +''
+        value = String.new(capacity: count + 1)
```
casperisfine, line 16: The `+ 1` is to account for the `\0` at the end of the String.

petergoldstein:
@casperisfine Good catch/optimization.  Thanks.

PR #939, opened 2022-11-29
Pass Encoding instances rather than encoding names as String

casperisfine:
This doesn't really matter much in the grand scheme of things but is a pet peeve of mine. It's very marginally faster as Ruby doesn't have to search for that name in the encoding table.


petergoldstein:
Sure, this is reasonable.  Thanks!

PR #942, opened 2022-12-10
interleave read and write on pipelined_get (#776, #941)

### LINKED ISSUE #776: send_multiget write call hangs if keys total bytesize > ~1,700,000 bytes
We are using dalli 2.7.11 with Sinatra 2.1.0 and Ruby 2.7.1, and an AWS Elasticache cluster of 3 `cache.m6g.2xlarge` machines.

We just migrated from the old memcached gem and had no problems.

Everything is working fine with smaller total keys bitesizes. 

**But when the total keys bytesize exceeds about 1,700,000 bytes, the `write(req)` call just hangs.**

We instrumented the dalli gem code and spent two days troubleshooting to come to this conclusion.

To test, we set up a sample key space and then tried running `multi_get` with different numbers of keys. It didn't matter if we had mostly cache misses, or if there was a small or large amount of data stored in each key.

```
keys = []
(0...1_000_000).each do |i|
    keys << "xyz0123:#{i}"
end

dalli_client = Dalli::Client.new(
    <addresses>,
    timeout: 0.5,
    error_when_over_max_size: true
)
dalli_client.get_multi(keys[0...124_000]).size
```

Stack Trace: (If we hit Ctrl-C to stop processing)

```
Traceback (most recent call last):
       16: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/client.rb:76:in `block in get_multi'
       15: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/client.rb:423:in `get_multi_yielder'
       14: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/client.rb:364:in `perform'
       13: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/client.rb:425:in `block in get_multi_yielder'
       12: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/ring.rb:52:in `lock'
       11: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/client.rb:431:in `block (2 levels) in get_multi_yielder'
       10: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/client.rb:305:in `make_multi_get_requests'
        9: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/client.rb:305:in `each'
        8: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/client.rb:311:in `block in make_multi_get_requests'
        7: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/options.rb:18:in `request'
        6: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/options.rb:18:in `synchronize'
        5: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/options.rb:19:in `block in request'
        4: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/server.rb:70:in `request'
        3: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/server.rb:288:in `send_multiget'
        2: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/server.rb:575:in `write'
        1: from /usr/local/lib/ruby/gems/2.7.0/gems/dalli-2.7.11/lib/dalli/server.rb:575:in `write'
```

Can anyone help?

### LINKED ISSUE #941: Hanging when performing read_multi against large number of keys
We're seeing an issue where the dalli client hangs trying to read ~100-200K keys at once. We were able to repro with the following code and a cache store configuration.

It looks similar to https://github.com/petergoldstein/dalli/issues/776 but wanted to confirm and potentially add our use case given the age of that issue. We're going to mitigate on our end by dropping the batch size to 10K keys per call.

* `ruby 2.7.6p219`
* `Rails 6.0.5.1`
* `dalli (3.2.3)`

```ruby
  config.cache_store = :mem_cache_store,
    ENV.fetch('MEMCACHE_SERVERS', '127.0.0.1:11211'),
    { :namespace => "comp-#{Rails.env}" }
```

```ruby
require 'timeout'

key_count = 2**10

while key_count < 10_000_000 do
    prefix = (rand * 2**64).to_i.to_s(32)
    puts "Prepping #{key_count} keys"
    keys = key_count.times.map { |i| "some_long_string:#{prefix}:#{i}" }
    t = Time.now
    puts "Starting set multi #{t}"
    Rails.cache.fetch_multi(*keys) { 0 }
    diff = Time.now - t
    puts "Completed in #{diff}s"
    t = Time.now
    puts "Starting multi get #{t}"
    Timeout::timeout(diff) {
        Rails.cache.read_multi(*keys) { 0 }
    }
    puts "Completed in #{Time.now - t}s"
    key_count *= 2
end
```

marvinthepa:
fixes #776.
fixes #941.

When reading a large number of keys, memcached starts sending the
response when dalli is not yet finished sending the request.

As we did not start reading the response until we were finished writing
the request, this could lead to the following problem:

* the receive buffer (rcvbuf) would fill up
* due to TCP backpressure, memcached would stop sending (okay as we are
  not reading anyway), but also stop reading
* the send buffer (sndbuf) would also fill up
* as we were using a blocking write without timeout, we would block
  forever (at least with ruby < 3.2, which introduces IO::Timeout, see
  #967)

This is addressed by using IO::select on the sockets for both read and
write, and thus start reading as soon as data is available.

### DIFF
--- a/lib/dalli/pipelined_getter.rb
+++ b/lib/dalli/pipelined_getter.rb
@@ -17,9 +17,8 @@ module Dalli
       return {} if keys.empty?
 
       @ring.lock do
-        servers = setup_requests(keys)
-        start_time = Time.now
-        servers = fetch_responses(servers, start_time, @ring.socket_timeout, &block) until servers.empty?
+        requests = setup_requests(keys)
+        fetch_responses(requests, @ring.socket_timeout, &block)
       end
     rescue NetworkError => e
       Dalli.logger.debug { e.inspect }
@@ -27,58 +26,17 @@ module Dalli
       retry
     end
 
-    def setup_requests(keys)
-      groups = groups_for_keys(keys)
-      make_getkq_requests(groups)
-
-      # TODO: How does this exit on a NetworkError
-      finish_queries(groups.keys)
-    end
-
-    ##
-    # Loop through the server-grouped sets of keys, writing
-    # the corresponding getkq requests to the appropriate servers
-    #
-    # It's worth noting that we could potentially reduce bytes
-    # on the wire by switching from getkq to getq, and using
-    # the opaque value to match requests to responses.
-    ##
-    def make_getkq_requests(groups)
-      groups.each do |server, keys_for_server|
-        server.request(:pipelined_get, keys_for_server)
-      rescue DalliError, NetworkError => e
-        Dalli.logger.debug { e.inspect }
-        Dalli.logger.debug { "unable to get keys for server #{server.name}" }
+    def setup_requests(all_keys)
+      groups_for_keys(all_keys).to_h do |server, keys|
+        # It's worth noting that we could potentially reduce bytes
+        # on the wire by switching from getkq to getq, and using
+        # the opaque value to match requests to responses.
+        [server, server.pipelined_get_request(keys)]
       end
-    end
-
-    ##
-    # This loops through the servers that have keys in
-    # our set, sending the noop to terminate the set of queries.
-    ##
-    def finish_queries(servers)
-      deleted = []
-
-      servers.each do |server|
-        next unless server.connected?
-
-        begin
-          finish_query_for_server(server)
-        rescue Dalli::NetworkError
-          raise
-        rescue Dalli::DalliError
-          deleted.append(server)
-        end
-      end
-
-      servers.delete_if { |server| deleted.include?(server) }
-    rescue Dalli::NetworkError
-      abort_without_timeout(servers)
-      raise
     end
 
     def finish_query_for_server(server)
-      server.pipeline_response_setup
+      server.finish_pipeline_request
     rescue Dalli::NetworkError
       raise
     rescue Dalli::DalliError => e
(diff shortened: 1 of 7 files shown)


lib/dalli/pipelined_getter.rb
```diff
@@ -92,29 +50,101 @@ def abort_without_timeout(servers)
...
+    rescue Dalli::NetworkError
       abort_without_timeout(servers)
       raise
+    rescue DalliError => e
+      Dalli.logger.debug { e.inspect }
+      Dalli.logger.debug { "unable to get keys for server #{server.name}" }
+    end
+
+    def server_pipelined_get(server, request)
+      buffer_size = server.socket_sndbuf
```
petergoldstein, line 137: I had been thinking about doing it by batching the number of keys, but I think this approach of looking it at from the perspective of # of bytes is a better way to address the issue.  It also allows us to give the end user direction about how to control the behavior (by tuning sndbuf)

lib/dalli/pipelined_getter.rb
```diff
@@ -92,29 +50,101 @@ def abort_without_timeout(servers)
...
+      # FIXME: this was here. why. where should it go?
       # Remove any servers which are not connected
-      servers.delete_if { |s| !s.connected? }
-      return [] if servers.empty?
+      # servers.delete_if { |s| !s.connected? }
+
+      start_time = Time.now
+      servers = requests.keys
+
+      # FIXME: this was executed before the finish request was sent. Why?
```
petergoldstein, line 61: I have to look at this in a little detail.  The behavior in the case of no servers available is supposed to a returned `[]` with no error.  But if I recall correctly, there are cases where the a server is not connected that can trigger raising a `Dalli::Error` without this check.  I'll find some time to look.

lib/dalli/protocol/base.rb
```diff
@@ -143,6 +145,14 @@ def quiet?
       end
       alias multi? quiet?
 
+      def pipelined_get_request(keys)
+        req = +''
```
petergoldstein, line 149: I know this line was just copied and moved from pre-existing code, but looking at it we can almost certainly reduce allocations by pre-allocating `req` to a reasonable size.  Something like `keys.size * 24 + keys.map(&:size).sum)` for binary? And something like `keys.size * 18 + keys.map(&:size).sum)` for meta.

test/integration/test_pipelined_get.rb
```diff
@@ -79,6 +79,27 @@
...
 
+      it 'does not block for a large number of existing keys' do
+        memcached_persistent(p) do |dc|
+          dc.close
+          dc.flush
+
+          key_count = 200_000
+          range = 0...key_count
+          dc.quiet do
+            range.each { dc.set(_1, "foobar_#{_1}") }
```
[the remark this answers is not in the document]
marvinthepa: Oh sorry, I'll fix that.

lib/dalli/pipelined_getter.rb
marvinthepa: I started out with batching the keys, but I got the idea to batch by bytes instead when I struggled to come up with a good default value for the batch size - It seemed to me that the original problem was caused by the response size rather than the number of keys requested..

I don't know if sndbuf is a good value though, maybe it is more efficient to try to write larger batches than the buffer.
It will also not allow users of the library to tweak the send buffer and the chunk size independently (I do not know if that makes sense).

It might not matter at all as this uses nonblocking IO for the writes, which could mean that there would just be partial writes (which are already correctly handled AFAIK) if we try to write everything at once. Maybe I will do some experiments, but I think there are other areas I should polish first (basic correctness being the most important one).

lib/dalli/pipelined_getter.rb
marvinthepa: Any hints on the intended original behavior of the multi-server code are highly appreciated. I haven't been able to fully wrap my head around it yet, and I guess it is the hardest to get right when things fail in between reading/writing chunks. Hard to correctly test as well..

lib/dalli/protocol/base.rb
marvinthepa: Will do. But it seems that it should be 17 for meta, at least it was in my tests (non-base64, which will completely break this calculation..).
Does not seem to make a significant difference in speed, though.

I have another idea for a performance optimization: start of by only generating as much bytes as are needed for the first chunk and send it, then pre-calculate upcoming bytes whenever the select blocks (or the socket is writable and no pre-calculated data is available).
This means the server can start generating the response earlier, dalli would use less memory, and time that is "lost" waiting for I/O could be used for calculation.
But it will be a bit harder to do elegantly without a library like async or eventmachine, and I don't know if it is worth the added complexity. Anyway, I would put this in an extra PR once this one is finished.

marvinthepa:
> Thanks for drafting @marvinthepa

Happy to give back and spend some company time on a library we have been using for a few years now..

lib/dalli/pipelined_getter.rb
petergoldstein: I think we might do something like this:

1. Add an explicit chunking parameter that can be set on the client
2. Have that chunking parameter fallback to the sndbuf size if it isn't explicitly set

lib/dalli/protocol/base.rb
petergoldstein: It's very possible I got the number wrong for meta (I did the calculation quickly) but just make sure you're including the separator.

I think the benefit for that optimization is unlikely to be worth the tradeoff, but we can discuss.

lib/dalli/protocol/base.rb
```diff
@@ -201,13 +211,14 @@ def connect
...
-        # Could send noop here instead of in pipeline_response_setup
-        write(req)
+      def pipelined_get(bytes)
+        write_nonblock(bytes)
+      rescue SystemCallError, Timeout::Error, EOFError => e
+        @connection_manager.error_on_request!(e)
+      end
+
+      def pipelined_get_capacity(keys)
+        (keys.size * request_header_size) + keys.reduce(0) { |acc, k| acc + k.size }
```
jamiemccarthy, line 221:
```suggestion
        (keys.size * request_header_size) + keys.map(&:size).sum
```

In some quick benchmarking this looks 12-15% faster maybe, and I think more readable

lib/dalli/protocol/base.rb
jamiemccarthy: I think that's a neat idea for an optimization, but yeah it feels like it could be an improvement for a future PR.

The `.map(&:size).sum)` is faster in my little benchmark, but by only by like 10 nanoseconds a key.

marvinthepa: > The `.map(&:size).sum)` is faster in my little benchmark, but by only by like 10 nanoseconds a key.

I had the (naive) assumption that the cost of allocating an array should be prevented - not knowing that it is outweighed by doing the sum in C instead if in Ruby. I should have remembered Knuth.

marvinthepa:
@petergoldstein sorry for not following up on this PR for so long, but I found some time and hope I can manage to get this ready to integrate in the next weeks.

I rebased on main, fixing the conflicts with especially #963 - so maybe if @casperisfine could check this PR as well it would be highly appreciated.

I am still not 100% confident if the error handling is completely correct, but I was not able to find a flaw (at least not one that did not exist before anyway) after some time searching for one.

lib/dalli/pipelined_getter.rb
marvinthepa: :heavy_check_mark: 

I hope you like the parameter-name `chunk_size`.

I would have updated the documentation as well, but as it is not in the repo but in the wiki I guess this has to wait until it is released..

marvinthepa:
Back to draft, as I think I found a problem relating to #963, where the connection is unnecessarily closed during a pipelined get. *sigh*.

PR #953, opened 2023-02-16
Cache Process.pid on Ruby 3.1+

casperisfine:
Ref: https://bugs.ruby-lang.org/issues/19443

Recent `glibc` no longer cache the PID, so every call to `Process.pid` end up emitting a syscall which as an impact on performance.

Ruby 3.3 may or may not do PID caching, in the meantime we can implement our own caching in Ruby 3.1+ using the `Process._fork` callback.

NB: This is very much the same code I'm about to merge in redis-client: https://github.com/redis-rb/redis-client/pull/91, and probably Active Record too https://github.com/rails/rails/pull/47419

FYI @dalehamel

### DIFF
--- a/lib/dalli/pid_cache.rb
+++ b/lib/dalli/pid_cache.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module Dalli
+  ##
+  # Dalli::PIDCache is a wrapper class for PID checking to avoid system calls when checking the PID.
+  ##
+  module PIDCache
+    if !Process.respond_to?(:fork) # JRuby or TruffleRuby
+      @pid = Process.pid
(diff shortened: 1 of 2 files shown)


lib/dalli/pid_cache.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Dalli
+  module PIDCache
```
petergoldstein, line 4: @casperisfine  Just add the below above the `module PIDCache` line and the problem should be resolved.  Once that's done I'll merge.

```
##
# Dalli::PIDCache is a wrapper class for PID checking to avoid system calls when checking the PID.
##
```

casperisfine: Done

PR #957, opened 2023-03-08
Prevent multiget socket corruption - fixes #956

### LINKED ISSUE #956: Race condition in get_multi leads to socket corruption and incorrect values returned for future gets
Hi, thanks for all your work maintaining this repo! I am a security engineer investigating some incorrect Rails caching behavior observed after upgrading to the latest version from 2.7.x, and have identified a mechanism consistent with my findings:

There is a race condition in the pipelined getter [implementation](https://github.com/petergoldstein/dalli/blob/e692421ebc923144a98716f079ac5e4d172815a0/lib/dalli/pipelined_getter.rb#L30-L36) which allows for socket corruption. The key problem is that a lock (`request_in_progress`) is acquired on the connection only after sending the getkq requests; if something happens between sending those requests and acquiring the connection lock, the socket could be left in a state which is both corrupt and (superficially) ready to accept future requests. 

In the corrupted state, subsequent operations which result in data being read from the socket will have incorrect behavior, because they expect that the next bytes on the socket are the responses to their own requests but the unread getkq responses will precede their own responses. **This could cause the client to return incorrect values for given keys because the writing and reading operations are out of sync / "off-by-x", where x is the number of responses to the getkq requests.**

### Suggested fix
I propose we should instead acquire a connection lock before writing the getkq requests to the socket. This means that if something happens after writing the getkq requests but before reading their responses, the connection will be left in the locked (`request_in_progress`) state and will refuse to accept additional requests (resetting instead). I am opening a PR which attempts to do this. 

### Open question
Note that reading the 2.7.x code, it seems that it could have had the same issue [here](https://github.com/petergoldstein/dalli/blob/884e1ff485df0adaff6e2095cd62cefb7badb34f/lib/dalli/client.rb#L424-L428). I am unsure if something else changed to make this failure case more likely - or if it was just a coincidence. The time window for a timeout/failure leading to this bug is narrow, so this  would certainly be a rare event with randomness at play; nonetheless I observed the incorrect behavior not long after upgrading to 3.2.3 which is suspicious and I would like to be able to explain what, if anything, changed.

### Demo/POC
in Rails console I used the following steps to reproduce this issue:
```
# code to produce a timeout on the 3.x Dalli gem
# simulates a process timeout occurring between make_getkq_requests and finish_queries
Dalli::PipelinedGetter.class_eval do 
  def setup_requests(keys)
    groups = groups_for_keys(keys)
    make_getkq_requests(groups)
    
    sleep 10 
    
    # TODO: How does this exit on a NetworkError
    finish_queries(groups.keys)
  end
end

c = ActiveSupport::Cache.lookup_store(:mem_cache_store)

# populate some cache entries by doing stuff like 
c.fetch("test10") { "test10result" }
c.fetch("test1") { "test1result1" }
c.fetch("test2") { "test1result2" }
c.fetch("test3") { "test1result3" }

# trigger timeout  during read_multi
Timeout.timeout(5) do 
  c.read_multi(*["test1", "test2", "test3", "test4"])
end

# this should now output the wrong result
c.fetch("test10")
```
=> "test1result1"

cornu-ammonis:
See https://github.com/petergoldstein/dalli/issues/956.

The key idea is to treat the entire series of write and read ops as a single "request" for the connection; because if we start the getkq write ops but do not read the results off the socket, the socket is left in an invalid state.

To test this locally in Rails Console I used the snippets referenced in the issue to contrive a timeout error. If you agree with this approach I'd be happy to work out how to test this in specs.

### DIFF
--- a/lib/dalli/options.rb
+++ b/lib/dalli/options.rb
@@ -31,6 +31,12 @@ module Dalli
       end
     end
 
+    def pipeline_get_setup 
+      @lock.synchronize do
+        super
+      end
+    end
+
     def pipeline_response_setup
       @lock.synchronize do
         super
--- a/lib/dalli/pipelined_getter.rb
+++ b/lib/dalli/pipelined_getter.rb
@@ -45,6 +45,7 @@ module Dalli
     ##
     def make_getkq_requests(groups)
       groups.each do |server, keys_for_server|
+        server.pipeline_get_setup
         server.request(:pipelined_get, keys_for_server)
       rescue DalliError, NetworkError => e
         Dalli.logger.debug { e.inspect }
--- a/lib/dalli/protocol/base.rb
+++ b/lib/dalli/protocol/base.rb
@@ -18,7 +18,7 @@ module Dalli
 
       def_delegators :@value_marshaller, :serializer, :compressor, :compression_min_size, :compress_by_default?
       def_delegators :@connection_manager, :name, :sock, :hostname, :port, :close, :connected?, :socket_timeout,
-                     :socket_type, :up!, :down!, :write, :reconnect_down_server?, :raise_down_error
+                     :socket_type, :up!, :down!, :write, :write_nonblock, :reconnect_down_server?, :raise_down_error
 
       def initialize(attribs, client_options = {})
         hostname, port, socket_type, @weight, user_creds = ServerConfigParser.parse(attribs)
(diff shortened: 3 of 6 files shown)


petergoldstein:
See my notes on #956.  Barring an updated example of the issue, I'll plan to close this in a couple of days.

Issue #967, opened 2023-07-19
IO::TimeoutError exceptions are not handled, breaking automatic retry


skaes:
In Ruby 3.2.0, sockets can raise IO::TimeoutError when blocking operations time out ([Ruby 3.2.0 News](https://github.com/ruby/ruby/blob/ruby_3_2/NEWS.md), [feature 18630](https://bugs.ruby-lang.org/issues/18630)) . These happen on read/write operations in Dalli. Dalli only rescues Timeout::Error.

The overall effect is that operations that time out are never retried. Instead they just log an error:

```
Unexpected exception during Dalli request: IO::TimeoutError: Blocking operation timed out!
```
I'm not sure why this happens, because from the Ruby NEWS file one would assume this only happens when the timeout attribute is set on the IO instance and I could not find it in the Dalli code.

skaes:
https://github.com/petergoldstein/dalli/pull/968 is an attempt to improve the situation.

petergoldstein:
This is merged and published in 3.2.6

skaes:
@petergoldstein would you mind tagging the release?

PR #968, opened 2023-07-20
rescue IO::TimeoutError raised by Ruby since 3.2.0 on blocking reads/writes

skaes:
One scenario in which this happens is when a firewall silently drops connections between clients and memcached instances.

With this patch the client reopens the connection and the request succeeds. Without it, IO:TimeoutError is raised for the application to handle which deviates from Dalli behavior under Ruby 3.1 and earlier.

There re several ways to patch this up. I opted for introducing an array of timeout classes for the rescue clauses.

### DIFF
--- a/lib/dalli/protocol.rb
+++ b/lib/dalli/protocol.rb
@@ -4,5 +4,14 @@ module Dalli
   module Protocol
     # Preserved for backwards compatibility.  Should be removed in 4.0
     NOT_FOUND = ::Dalli::NOT_FOUND
+
+    # Ruby 3.2 raises IO::TimeoutError on blocking reads/writes, but
+    # it is not defined in earlier Ruby versions.
+    require 'timeout'
+    if defined?(IO::TimeoutError)
+      TIMEOUT_ERRORS = [Timeout::Error, IO::TimeoutError]
+    else
+      TIMEOUT_ERRORS = [Timeout::Error]
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


skaes:
See https://github.com/petergoldstein/dalli/issues/967

skaes:
Another option to fix this would be to use IOError in the rescue clauses, as IO::TimeoutError is a subclass of IOError.

petergoldstein:
@skaes This looks ok to me, but it needs a rebase and a spec run.  Can you please rebase the branch and we'll get CI running?

Thanks.

skaes:
@petergoldstein the pull request was created from https://github.com/petergoldstein/dalli/commit/b69b24e943b2c7f5da688c80ee83bb7d7ee264fb which is the last commit on main. I don't know how to rebase it even more.

But really happy this gets addressed.

petergoldstein:
@skaes You were correct - there was no rebase required when I made my last comment.  But I just made some commits (primarily to address test failures).  Would you be able to rebase now and we can get specs running?

Sorry about the confusion.

skaes:
@petergoldstein rebased.

petergoldstein:
@skaes Thanks for rebasing.

Can you please address the lints identified by Rubocop and get Rubocop passing?  Then we can look at merging.

Thanks.

skaes:
I fixed what I could. I left the task of comment what module Protocol does to you, as I did not create that module.

petergoldstein:
Merged in #973 with the final lint fixed.  Thanks @skaes !

PR #973, opened 2023-09-25
rescue IO::TimeoutError raised by Ruby since 3.2.0 on blocking reads/writes

petergoldstein:
One scenario in which this happens is when a firewall silently drops connections between clients and memcached instances.

With this patch the client reopens the connection and the request succeeds. Without it, IO:TimeoutError is raised for the application to handle which deviates from Dalli behavior under Ruby 3.1 and earlier.

Addresses #967 and fixes the remaining lint in #968

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Dalli Changelog
 Unreleased
 ==========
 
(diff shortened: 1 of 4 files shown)


petergoldstein:
Thanks @skaes !

PR #982, opened 2024-01-23
Add base64 to gemspec

### LINKED ISSUE #981: Ruby 3.3: warning related to base64 missing in the gemspec
```
~/.rvm/gems/ruby-3.3.0/gems/dalli-3.2.6/lib/dalli/protocol/meta/key_regularizer.rb:3: warning: base64 was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec. Also contact author of dalli-3.2.6 to add base64 into its gemspec.
```

Going to submit a PR

tagliala:
base64 is loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0.

Adding `base64` to gemspec to avoid this warning in Ruby 3.3 and allow Dalli to run against ruby-head

Close #981

### DIFF
--- a/dalli.gemspec
+++ b/dalli.gemspec
@@ -22,4 +22,6 @@ Gem::Specification.new do |s|
   s.metadata = {
     'rubygems_mfa_required' => 'true'
   }
+
+  s.add_dependency 'base64'
 end


petergoldstein:
Thanks @tagliala !

Issue #966, opened 2023-07-10
No request in progress


jaredmoody:
I'm seeing this error pop up in my logs intermittently, always with this jbuilder template:

```
json.array! @projects, partial: 'project', as: :project, cached: ->(project) { [project.location, project] }
```

Recently, this template was changed, and the error started after changing `cached: true` to what it is above.

Abbreviated stacktrace:

```
vendor/bundle/ruby/3.0.0/gems/dalli-3.2.5/lib/dalli/protocol/connection_manager.rb:107:in `confirm_in_progress!': [Dalli] No request in progress. This may be a bug in Dalli. (ActionView::Template::Error)
    from vendor/bundle/ruby/3.0.0/gems/dalli-3.2.5/lib/dalli/protocol/base.rb:179:in `verify_pipelined_state'
    from vendor/bundle/ruby/3.0.0/gems/dalli-3.2.5/lib/dalli/protocol/base.rb:75:in `pipeline_response_setup'
    from vendor/bundle/ruby/3.0.0/gems/dalli-3.2.5/lib/dalli/pipelined_getter.rb:81:in `finish_query_for_server'
    from vendor/bundle/ruby/3.0.0/gems/dalli-3.2.5/lib/dalli/pipelined_getter.rb:66:in `block in finish_queries'
    from vendor/bundle/ruby/3.0.0/gems/dalli-3.2.5/lib/dalli/pipelined_getter.rb:62:in `each'
    from vendor/bundle/ruby/3.0.0/gems/dalli-3.2.5/lib/dalli/pipelined_getter.rb:62:in `finish_queries'
    from vendor/bundle/ruby/3.0.0/gems/dalli-3.2.5/lib/dalli/pipelined_getter.rb:35:in `setup_requests'
    from vendor/bundle/ruby/3.0.0/gems/dalli-3.2.5/lib/dalli/pipelined_getter.rb:20:in `block in process'
    from vendor/bundle/ruby/3.0.0/gems/dalli-3.2.5/lib/dalli/ring.rb:78:in `lock'
    from vendor/bundle/ruby/3.0.0/gems/dalli-3.2.5/lib/dalli/pipelined_getter.rb:19:in `process'
    from vendor/bundle/ruby/3.0.0/gems/dalli-3.2.5/lib/dalli/client.rb:108:in `block in get_multi'
    from vendor/bundle/ruby/3.0.0/gems/dalli-3.2.5/lib/dalli/client.rb:107:in `get_multi'
[...]

PR #983, opened 2024-01-25
Don't reconnect to send pipelined request no-op

eugeneius:
Fixes https://github.com/petergoldstein/dalli/issues/966.

If sending the pipelined get commands fails with a network error, the socket will be closed when we try to send the no-op command. `alive?` reconnects if possible, but we only need to send the no-op if the get commands were successfully sent.

Before https://github.com/petergoldstein/dalli/pull/963, reconnecting to send a no-op was harmless, but now that pipelined requests are tracked in their entirety, an error is raised since there is no request in progress.

### DIFF
--- a/lib/dalli/pipelined_getter.rb
+++ b/lib/dalli/pipelined_getter.rb
@@ -60,7 +60,7 @@ module Dalli
       deleted = []
 
       servers.each do |server|
-        next unless server.alive?
+        next unless server.connected?
 
         begin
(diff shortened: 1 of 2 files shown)


petergoldstein:
Thanks @eugeneius .  Rebased to address some test/rubocop issues.  If this runs green I'll merge.

petergoldstein:
Good catch here.

PR #986, opened 2024-01-29
Drop dependency on base64

Earlopain:
#982 added `base64` as a dependency in order to avoid a warning with Ruby 3.3

The `base64` gem is mostly a wrapper around `pack`/`unpack` and as such is easy to replace. Other gems have opted for the same approach:
* https://github.com/rack/rack/pull/2110
* https://github.com/lostisland/faraday/pull/1541
* https://github.com/rubocop/rubocop/pull/12313
* https://github.com/sidekiq/sidekiq/pull/6151
* etc

### DIFF
--- a/dalli.gemspec
+++ b/dalli.gemspec
@@ -22,6 +22,4 @@ Gem::Specification.new do |s|
   s.metadata = {
     'rubygems_mfa_required' => 'true'
   }
-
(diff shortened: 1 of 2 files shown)


petergoldstein:
I like this approach.

petergoldstein:
Thanks @Earlopain

Issue #987, opened 2024-02-01
TypeError: no implicit conversion of Hash into String


denisj:
Dalli 3.2.7
Ruby 3.2.3
Rails 7.1.3

I get this error when trying simple command such as
```ruby
::Rails.cache.fetch("gh:ec9:v1")
```

in lib/dalli/socket.rb line 92
```ruby
def self.open(host, port, options = {})
  sock = new(host, port, connect_timeout: options[:socket_timeout])
  sock.options = { host: host, port: port }.merge(options)
  init_socket_options(sock, options)

  options[:ssl_context] ? wrapping_ssl_socket(sock, host, options[:ssl_context]) : sock
end
```

in lib/ruby/3.2.0/resolv-replace.rb line 24
```ruby
class TCPSocket < IPSocket
 # :stopdoc:
  alias original_resolv_initialize initialize
  # :startdoc:
  def initialize(host, serv, *rest)
    rest[0] = IPSocket.getaddress(rest[0]) if rest[0]
    original_resolv_initialize(IPSocket.getaddress(host), serv, *rest)
  end
end
```

The thing present in rest[0] is
```ruby
[{:connect_timeout=>1}]
```

mem_cache_store configuration:
```ruby
config.cache_store = :mem_cache_store,
  '127.0.0.1:11211',
  { namespace: 'conversation', compress: true }
```

Backtrace
```
Feb 01 09:27:29 nostromo-manager-1 sidekiq[3979401]: 2024-02-01T09:27:29.066Z pid=3979401 tid=2g73h WARN: TypeError: no implicit conversion of Hash into String
Feb 01 09:27:29 nostromo-manager-1 sidekiq[3979401]: 2024-02-01T09:27:29.066Z pid=3979401 tid=2g73h WARN: /usr/share/rvm/rubies/ruby-3.2.3/lib/ruby/3.2.0/resolv.rb:110:in `each_address'
Feb 01 09:27:29 nostromo-manager-1 sidekiq[3979401]: /usr/share/rvm/rubies/ruby-3.2.3/lib/ruby/3.2.0/resolv.rb:93:in `getaddress'
Feb 01 09:27:29 nostromo-manager-1 sidekiq[3979401]: /usr/share/rvm/rubies/ruby-3.2.3/lib/ruby/3.2.0/resolv.rb:44:in `getaddress'
[...]

PR #991, opened 2024-02-06
Move check whether TCPSocket was patched to TCP.open

y9v:
This is a follow-up for #987.

We cannot check if `resolv-replace` changed `TCPSocket#initialize` method, since we have no guarantee that `resolv-replace` will be required before `dalli`. This PR moves this check to `TCP.open` method.


petergoldstein:
Thanks!

PR #993, opened 2024-02-08
Handle IO::TimeoutError when establishing connection

eugeneius:
Since https://github.com/petergoldstein/dalli/pull/977, `Dalli::Socket::TCP.open` can also raise this error.


petergoldstein:
Thanks @eugeneius !

PR #989, opened 2024-02-02
Fix conflict with resolv-replace gem

y9v:
Closes #987

In version 3.2.7 socket_timeout option was introduced for TCPSocket.

This works unless `resolv-replace` gem is loaded (which was added to ruby standard library since version 3.0.0).

This commit adds another check besides the ruby version check to avoid breaking dalli for applications that have `resolv-replace` gem required.

PR #998, opened 2024-03-07
Add `resolv-replace` compatibility test

sambostock:
While `dalli` doesn't depend on `resolv-replace`, we want to ensure it still works if it is required by the host application. (#989)

This regression test ensures we don't break compatibility, and adds a framework allowing us to test compatibility with other gems in the future.


test/test_gem_compatibility.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+# JRuby does not support forking, and it doesn't seem worth the effort to make it work.
+return unless Process.respond_to?(:fork)
```
sambostock, line 4: Active Support has some [workarounds](https://github.com/rails/rails/blob/608c1bfe17a2626304b6aae96a3824517fc07bb2/activesupport/lib/active_support/testing/isolation.rb) for this we could use, but it doesn't seem worth it for a code path that was previously untested, and is unlikely to behave differently on JRuby.

test/test_gem_compatibility.rb
```diff
@@ -0,0 +1,87 @@
...
+
+require_relative 'helper'
+
+describe 'gem compatibility' do
+  %w[
+    resolv-replace
+  ].each do |gem_name|
+    it "passes smoke test with #{gem_name.inspect} gem required" do
+      memcached(:binary, rand(21_397..21_896)) do |_, port|
+        in_isolation(timeout: 10) do
```
sambostock, line 14: We want to be able to require the gem in isolation so as not to affect the host process and other tests.

test/test_gem_compatibility.rb
```diff
@@ -0,0 +1,87 @@
...
+    expected = SecureRandom.hex(4)
+    key = "round-trip-#{expected}"
+    ttl = 10 # seconds
+
+    client.set(key, expected, ttl)
+
+    assert_equal(expected, client.get(key), message)
+  end
+
+  def in_isolation(timeout:) # rubocop:disable Metrics
```
sambostock, line 41: This method triggers several offences in the `Metrics` department, but I don't know that splitting it would be an improvement.

petergoldstein:
Thanks @sambostock .  Cherry picked into 4.3.1

PR #999, opened 2024-03-08
Fix `socksify` incompatibility

### LINKED ISSUE #996: TypeError raised in latest Socksify when initializing Dalli client with version 3.2.7+
3.2.7 and 3.2.8 cause a TypeError in Socksify.  3.2.6 works OK.

Reproduction script, passes with socksify 1.7.1 and dalli 3.2.6 installed, fails with dalli 3.2.7+:

```ruby
require 'dalli'
dc = Dalli::Client.new('localhost:11211')
dc.set('abc', 'required dalli')
puts dc.get('abc')

require 'socksify'
dc = Dalli::Client.new('localhost:11211')
dc.set('xyz', 'required socksify')
puts dc.get('xyz')
```

Backtrace:

```
/home/.rvm/gems/ruby-3.0.6/gems/socksify-1.7.1/lib/socksify.rb:178:in `initialize': no implicit conversion of Hash into String (TypeError)
	from /home/.rvm/gems/ruby-3.0.6/gems/socksify-1.7.1/lib/socksify.rb:178:in `initialize'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/socket.rb:93:in `new'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/socket.rb:93:in `open'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/protocol/connection_manager.rb:210:in `memcached_socket'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/protocol/connection_manager.rb:55:in `establish_connection'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/protocol/base.rb:207:in `connect'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/protocol/base.rb:196:in `ensure_connected!'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/protocol/base.rb:58:in `alive?'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/options.rb:24:in `block in alive?'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/options.rb:23:in `synchronize'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/options.rb:23:in `alive?'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/ring.rb:46:in `server_for_key'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/client.rb:425:in `perform'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/client.rb:208:in `set_cas'
	from /home/.rvm/gems/ruby-3.0.6/gems/dalli-3.2.7/lib/dalli/client.rb:201:in `set'
	from ./typeerror.rb:8:in `<main>'
```

sambostock:
Similarly to `resolv-replace`, `sockify` monkey patches `TCPSocket#initialize` which causes an error if it is called with `connect_timeout:`.

Rather than check only for `resolv-replace`'s specific method, we can check if if the method's parameters have changed.

This PR builds on #998 and adds the `socksify` gem to the list of gems to smoke test compatibility with.

Closes #996

### ⚠️ Before Merging
- [ ] Rebase after #998 is merged


lib/dalli/socket.rb
```diff
@@ -97,15 +97,14 @@ def self.open(host, port, options = {})
         end
       end
 
+      TCPSOCKET_ORIGINAL_INITIALIZE_PARAMETERS = [[:rest].freeze].freeze
```
sambostock, line 100: This is memoized so we don't create duplicate arrays every time.

lib/dalli/socket.rb
```diff
@@ -97,15 +97,14 @@ def self.open(host, port, options = {})
...
-        # (part of ruby standard library since 3.0.0, should be removed in 3.4.0),
-        # as it does not handle keyword arguments correctly.
-        # To check this we are using the fact that resolv-replace
-        # aliases TCPSocket#initialize method to #original_resolv_initialize.
-        # https://github.com/ruby/resolv-replace/blob/v0.1.1/lib/resolv-replace.rb#L21
+        # Some gems like resolv-replace and socksify monkey patch TCPSocket and make it impossible to use the
+        # Ruby 3+ `connect_timeout:` keyword argument. If that's the case, we fallback to using the `Timeout` module.
         if RUBY_VERSION >= '3.0' &&
-           !::TCPSocket.private_instance_methods.include?(:original_resolv_initialize)
+           ::TCPSocket.instance_method(:initialize).parameters == TCPSOCKET_ORIGINAL_INITIALIZE_PARAMETERS
```
sambostock, line 107: This just checks for the original parameters, but we could also use `connect_timeout` if we detect any of
- `parameters.include?([:key, :connect_timeout])`
- `parameters.include?([:keyreq, :connect_timeout])`
- `parameters.any? { |type, _name| type == :keyrest }`

I kept it simple, as these would increase the runtime penalty, and I'm not aware of any gems that do monkey patch `TCPSocket#initialize` correctly, nor am I certain we _must_ use `connect_timeout:` instead of `Timeout` in those cases. 🤷

sambostock (commented):
The first commit is from #998, which I made as a separate PR because I think it makes sense to have the compatibility tests even if we decide to simply go back to exclusively relying on `Timeout` instead of `connect_timeout:`, as was discussed on https://github.com/petergoldstein/dalli/pull/989#issuecomment-1951384728.

sambostock:
I see

> ```
> failed to listen on TCP port 21345: Address already in use
> failed to listen on TCP port 33044: Address already in use
> failed to listen on TCP port 33044: Address already in use
> ```

in the [output of the only test runner that failed for the **first** commit](https://github.com/petergoldstein/dalli/actions/runs/8199040230/job/22423532447?pr=999#step:5:13), and similarly

> ```
> failed to listen on TCP port 21345: Address already in use
> ```

in the [output of the only test runner that failed for the **second** commit](https://github.com/petergoldstein/dalli/actions/runs/8199166353/job/22423826157?pr=999#step:5:13).

I feel like I've got flakey tests due to a bunch of the tests and helpers using hard coded ports instead of grabbing a free or simply random port.

PR #1000, opened 2024-03-08
Dynamically choose available ports in tests

sambostock:
Many tests appear to be flakey due to hard coded ports already being taken.

This refactors all ports to be chosen dynamically by checking for an available port first.


petergoldstein:
Thanks @sambostock , but as the underlying issue isn't a problem at this time I'm going to close the PR.

PR #1001, opened 2024-04-05
Only flush multi responses once

naveg:
`Client#quiet` is reentrant, in that multiple `quiet` blocks can be
nested. We only need to flush the responses at the conclusion of the
outermost block.


naveg:
An even better solution would be to address #1002 , so I'm closing this

PR #1004, opened 2024-06-04
Add a space to properly format multi-line log output

kevinebaugh:
🤓 

Current output:
```
Expiration interval (6569244) too long for Memcached and too short to be a future timestamp,converting to an expiration timestamp
```

With this fix:
```
Expiration interval (6569244) too long for Memcached and too short to be a future timestamp, converting to an expiration timestamp
```


erichkist (approved):

petergoldstein:
Thanks!

PR #1005, opened 2024-06-13
Add host_to_name_map option to Dalli::Protocol::ConnectionManager

philister:
New config-parameter: `host_to_name_map`
 a hash mapping "host:port" to a name. Useful for providing more descriptive names or  tuning the ring-continuum e.g. `{ 'localhost:11211' => 'memcached1', 'mem2.cach.ed' => 'memcached2'}`

For a consistent calculation of which key can be found on which server, the config must also be the same on all clients. Previously, this was only possible if all memcached servers were accessible at the same address on all clients. 
However, you may want or be able to reach the memcache servers from different clients only via different addresses (local vs public vs docker vs dns, etc)


lib/dalli/protocol/connection_manager.rb
```diff
@@ -45,7 +47,13 @@ def name
         if socket_type == :unix
```
grcooper, line 47: Why are we not also doing this for unix sockets?

lib/dalli/protocol/connection_manager.rb
```diff
@@ -45,7 +47,13 @@ def name
...
           hostname
         else
-          "#{hostname}:#{port}"
+          name = "#{hostname}:#{port}"
+
+          if options[:host_to_name_map][name].nil?
```
grcooper, line 52: This can be simplified to:

`options[:host_to_name_map][name] || name`

grcooper (commented):
This name method is buried in the library behind a collection of private methods. Ie to fetch it I believe you have to do things like use the `send` method? So I am curious as to what you want to do with "tuning the ring-continuum", as that is locked inside of Dalli itself generally.

```
c = Dalli::Client.new

c.send(:ring).servers.name # needing to do send is a code smell
```

I also know it's used in `stats` and `version` as well, but if you are returning this data in your app I assume you have some kind of name to server mapping that already exists and could map the results as needed?

I just don't want to add more complexity to the top-level options if unnecessary.

philister:
I am fine with closing the PR. This is not something we necessarily need in upstream.

PR #1010, opened 2024-08-08
Add logger as a dependency, drop ostruct for tests

olleolleolle:
This PR avoids a Ruby warning.

There was a warning in ruby-head (3.4) that in Ruby 3.5 `logger` and `ostruct` will not be built-in gems, but regular gems.

```
/home/runner/work/dalli/dalli/test/helper.rb:13: warning: logger was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.5.0. Add logger to your Gemfile or gemspec.
/home/runner/work/dalli/dalli/test/helper.rb:14: warning: ostruct was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.5.0. Add ostruct to your Gemfile or gemspec.
```

This change avoids that warning.

---

In addition, the `ostruct` dependency wasn't in actual use, just `require 'ostruct'` statements. It was possible to fully omit that and pass all tests.

A second commit with a linting change was added, using the autocorrect feature of RuboCop.

### DIFF
--- a/dalli.gemspec
+++ b/dalli.gemspec
@@ -24,4 +24,6 @@ Gem::Specification.new do |s|
     'changelog_uri' => 'https://github.com/petergoldstein/dalli/blob/main/CHANGELOG.md',
     'rubygems_mfa_required' => 'true'
   }
+
+  s.add_dependency 'logger'
 end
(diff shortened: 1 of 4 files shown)


petergoldstein:
@olleolleolle can you please address the Rubocop warnings?  Thanks.

olleolleolle:
@petergoldstein Thanks, I revisited the change, and was able to remove the ostruct, _and_ pass linting.

PR #1009, opened 2024-08-08
Refactor regex to be built using Regexp.union

olleolleolle:
This PR clarifies which error message strings are being matched, by using [Ruby's Regexp.union method](https://ruby-doc.org/core-2.4.2/Regexp.html) to build the patterns.

PR #1011, opened 2024-08-11
Serializer: reraise all .load errors as UnmarshalError

olleolleolle:
> Doesn’t _any_ error during deserialising mean the data is bad?

In order to avoid missing an error message to filter out, treat any Marshal.load error as a failed serialization, and trust Ruby's e.cause system (edit: introduced in Ruby 2.1) to provide a lineage of the error's true beginning.

This change was inspired by a comment relating to #1008.

With this rescue, it's possible to dismantle the other regular expressions.

---

This PR supersedes #1009, #1008


lib/dalli/protocol/value_serializer.rb
```diff
@@ -33,42 +33,17 @@ def store(value, req_options, bitflags)
...
-
-      def filter_name_error(err)
-        raise err unless err.message.include?(NAME_ERR_STR)
-
-        raise UnmarshalError, "Unable to unmarshal value: #{err.message}"
+        if serialized
+          begin
+            serializer.load(value)
+          rescue StandardError => e
+            raise UnmarshalError, "Unable to unmarshal value: #{e.message}"
```
ioquatix, line 42:
```suggestion
          rescue => error
            raise UnmarshalError, "Unable to unmarshal value!"
```

Cause will already carry the message, no point repeating it.

olleolleolle: (The `StandardError` mention is a nod to rubocop rules.)

ioquatix: Actually we don't even need `=> error` if we are just raising a more general error :)

olleolleolle: 🧠 == 🧠 (I pushed new.)

lib/dalli/protocol/value_serializer.rb
```diff
@@ -33,55 +33,17 @@ def store(value, req_options, bitflags)
...
-            begin
-              serializer.load(value)
-            rescue StandardError => e
-              raise UnmarshalError, "Unable to unmarshal value: #{e.message}"
-            end
-          else
-            # Use Dalli's existing exception filtering for deserialization when not using Marshal to serialize.
+          begin
             serializer.load(value)
+          rescue StandardError
```
ioquatix, line 41:
```suggestion
          rescue
```

This is identical (the default is `StandardError`).

olleolleolle: Yes, _and_, Rubocop rules in this project say

```

Offenses:

lib/dalli/protocol/value_serializer.rb:41:11: C: [Correctable] Style/RescueStandardError: Avoid rescuing without specifying an error class.
          rescue
          ^^^^^^
```

ioquatix: Haha never mind then :)

petergoldstein:
@olleolleolle So is this an alternative to #1008 ?  

Are there any concerns on how this works with older Ruby versions?  Will they see a meaningful change in behavior?

petergoldstein:
@olleolleolle It would helpful to understand how these different PRs relate, since they seem to touch similar code.

ioquatix (approved):
This is the best option by far. `cause` is supported as far back as 2.1 (https://docs.ruby-lang.org/en/2.1.0/Exception.html#method-i-cause) so I don't think depending on it for debugging failures is a problem, and in general, this PR should be much safer, since the alternative is unexpected and thus unhandled exceptions bubbling up through cache implementations (ultimately resulting in 🔥 in production).

olleolleolle:
> @olleolleolle It would helpful to understand how these different PRs relate, since they seem to touch similar code.

Ah, thanks for asking, they're sort of the evolution of my thinking on this. 

This is the more thorough change, relying on the Ruby feature of "cause" in exception handling.

The previous ones were about "not editing too much in the project", trying to use a light touch.

PR #1018, opened 2024-12-11
connection_manager: remove unnecessary variable

pjambet:
The `detail` variable was created twice, it seems to have been an artifact of a previous refactor


petergoldstein:
Thanks!

PR #1024, opened 2025-01-31
Handle SERVER_ERROR from Memcached

grcooper:
As per the Memcached spec, SERVER_ERROR is a valid value to return, alongside the error message. This only seems to be valid for the Meta/Text protocols.

I have added support so things like Envoy that return SERVER_ERROR will be handled correctly and raised as an appropriate error.

I also added the `debug` gem to `:development` and `:test` since it is very useful to be able to add things like breakpoints while debugging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Unreleased
 ==========
 
 - Fix cannot read response data included terminator `\r\n` when use meta protocol (matsubara0507)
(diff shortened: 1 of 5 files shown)


danmayer:
This looks good can you update the changelog, `Unreleased` section with a note on the fix.

https://github.com/petergoldstein/dalli/blob/main/CHANGELOG.md

if you have a good link to where it is mentioned in the memcached spec that would be nice, but I know memcached docs are all being updated and moving at the moment.

danmayer (changes requested):
small request on changelog

petergoldstein (approved):
This looks good to me.

PR #1025, opened 2025-02-03
feat: I/O Timeout

nickamorim:
## Description

Currently we're relying on the `SO_SNDTIMEO` and `SO_RCVTIMEO` socket options for connect/read/write timeouts. [As of CRuby 3.0, sockets are non-blocking by default](https://github.com/ruby/ruby/blame/78f188524f551c97b1a7a44ae13514729f1a21c7/ext/socket/init.c#L411-L434) which means that these socket options do not have any effect by default:

>Timeouts only have effect for system calls that perform socket I/O (e.g., [accept(2)](https://man7.org/linux/man-pages/man2/accept.2.html), [connect(2)](https://man7.org/linux/man-pages/man2/connect.2.html), [read(2)](https://man7.org/linux/man-pages/man2/read.2.html), [recvmsg(2)](https://man7.org/linux/man-pages/man2/recvmsg.2.html), [send(2)](https://man7.org/linux/man-pages/man2/send.2.html), [sendmsg(2)](https://man7.org/linux/man-pages/man2/sendmsg.2.html)); timeouts have no effect for [select(2)](https://man7.org/linux/man-pages/man2/select.2.html), [poll(2)](https://man7.org/linux/man-pages/man2/poll.2.html), [epoll_wait(2)](https://man7.org/linux/man-pages/man2/epoll_wait.2.html), and so on.

https://man7.org/linux/man-pages/man7/socket.7.html

This is easily testable by using [Toxiproxy](https://github.com/Shopify/toxiproxy) to inject latency to an upstream Memcached:
1. Follow the instructions to install Toxiproxy: https://github.com/Shopify/toxiproxy?tab=readme-ov-file#1-installing-toxiproxy
2. `memcached -d`
3. `toxiproxy-cli create -l localhost:22122 -u localhost:11211 memcached_latency`
4. Inject 5s of latency: `toxiproxy-cli toxic add -t latency -a latency=5000`
5. Create a Dalli client with a 50ms timeout: `c = Dalli::Client.new("127.0.0.1:22122", socket_timeout: 0.05)` 
6. Set a key: `c.set("foo", "bar", 0, raw: true)` and notice that it hangs for 10s (5s for the initial `version` call when establishing a connection and 5s for the `set` operation)

Then checkout `nickamorim/io-timeout` and you should get the following error:
```
W, [2024-11-11T13:18:10.996606 #84852]  WARN -- : 127.0.0.1:22122 failed (count: 0) IO::TimeoutError: Blocking operation timed out!
W, [2024-11-11T13:18:10.996684 #84852]  WARN -- : 127.0.0.1:22122 is down
dalli/lib/dalli/ring.rb:48:in `server_for_key': No server available (Dalli::RingError)
	dalli/lib/dalli/client.rb:441:in `perform'
	dalli/lib/dalli/client.rb:224:in `set_cas'
	dalli/lib/dalli/client.rb:217:in `set'
```

## Proposed Solution

This PR uses the `IO#timeout` feature added in https://bugs.ruby-lang.org/issues/18630 to set a timeout for read, write, and connect calls.

Note: I understand that it's non-ideal to have a singular timeout for read, write, and connect calls. I can fix this in a follow-up PR.

### DIFF
--- a/lib/dalli/socket.rb
+++ b/lib/dalli/socket.rb
@@ -117,19 +117,33 @@ module Dalli
       end
 
       def self.init_socket_options(sock, options)
+        configure_tcp_options(sock, options)
+        configure_socket_buffers(sock, options)
+        configure_timeout(sock, options)
+      end
+
+      def self.configure_tcp_options(sock, options)
         sock.setsockopt(::Socket::IPPROTO_TCP, ::Socket::TCP_NODELAY, true)
         sock.setsockopt(::Socket::SOL_SOCKET, ::Socket::SO_KEEPALIVE, true) if options[:keepalive]
+      end
+
+      def self.configure_socket_buffers(sock, options)
         sock.setsockopt(::Socket::SOL_SOCKET, ::Socket::SO_RCVBUF, options[:rcvbuf]) if options[:rcvbuf]
         sock.setsockopt(::Socket::SOL_SOCKET, ::Socket::SO_SNDBUF, options[:sndbuf]) if options[:sndbuf]
+      end
 
+      def self.configure_timeout(sock, options)
         return unless options[:socket_timeout]
 
-        seconds, fractional = options[:socket_timeout].divmod(1)
-        microseconds = fractional * 1_000_000
-        timeval = [seconds, microseconds].pack('l_2')
+        if sock.respond_to?(:timeout=)
+          sock.timeout = options[:socket_timeout]
+        else
+          seconds, fractional = options[:socket_timeout].divmod(1)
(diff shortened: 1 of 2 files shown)


grcooper (approved):

petergoldstein (changes requested):
Please address the Rubocop failures.  Otherwise this looks fine.

nickamorim:
> Please address the Rubocop failures.

I've been finding that the Rubocop rules are quite restrictive. Nonetheless they're now passing but we should revisit them.

danmayer (approved):

petergoldstein (approved):

PR #1029, opened 2025-02-12
update benchmark & profile to support meta client and commit performance fix for meta protocol

danmayer:
Ok, with the benchmark and profile we can start to pull over some performance improvements. I updated the benchmark and profile to support the meta protocol client as well as the binary. We haven't been working on the binary, but it is good to have both as we don't want to accidentally cause any regressions with the binary protocol support before we deprecate it.

dalli_string_profile
> using vernier profiler on the set command we can see it is spending a lot of time with string manipulation, this can be especially bad with very large values.

The before benchmark, shows that meta set is the slowest 
```
❯❯❯$ BENCH_TARGET=set RUBY_YJIT_ENABLE=1 bundle exec bin/benchmark                                                 <bundler> [main 
yjit: true
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) +YJIT [arm64-darwin23]
          client set   730.000 i/100ms
     meta client set   509.000 i/100ms
        raw sock set     1.221k i/100ms
          client set      6.279k (±10.3%) i/s -     62.780k in  10.101866s
     meta client set      4.547k (±13.4%) i/s -     44.792k in  10.012413s
        raw sock set     13.935k (± 2.2%) i/s -    140.415k in  10.082106s

Comparison:
        raw sock set:    13934.8 i/s
          client set:     6278.6 i/s - 2.22x  slower
     meta client set:     4546.7 i/s - 3.06x  slower
```

after benchmark, shows meta is now nearly as fast as raw socket
```
❯❯❯$ BENCH_TARGET=set RUBY_YJIT_ENABLE=1 bundle exec bin/benchmark                                                 <bundler> 
yjit: true
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) +YJIT [arm64-darwin23]
          client set   303.000 i/100ms
     meta client set     1.025k i/100ms
        raw sock set     1.436k i/100ms
          client set      6.573k (±10.8%) i/s -     64.842k in  10.004278s
     meta client set     13.347k (± 6.2%) i/s -    133.250k in  10.024490s
        raw sock set     14.580k (± 5.2%) i/s -    146.472k in  10.073444s

Comparison:
        raw sock set:    14580.3 i/s
     meta client set:    13347.1 i/s - same-ish: difference falls within error
          client set:     6573.4 i/s - 2.22x  slower
```

### DIFF
--- a/lib/dalli/protocol/meta.rb
+++ b/lib/dalli/protocol/meta.rb
@@ -85,6 +85,8 @@ module Dalli
                                         bitflags: bitflags, cas: cas,
                                         ttl: ttl, mode: mode, quiet: quiet?, base64: base64)
         write(req)
+        write(value)
+        write(TERMINATOR)
       end
       # rubocop:enable Metrics/ParameterLists
 
@@ -105,6 +107,8 @@ module Dalli
         req = RequestFormatter.meta_set(key: encoded_key, value: value, base64: base64,
                                         cas: cas, ttl: ttl, mode: mode, quiet: quiet?)
         write(req)
+        write(value)
+        write(TERMINATOR)
       end
       # rubocop:enable Metrics/ParameterLists
 
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Set up Ruby
       uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 3.2
+        ruby-version: 3.4
         bundler-cache: true # 'bundle install' and cache
     - name: Run Benchmarks
       run: RUBY_YJIT_ENABLE=1 BENCH_TARGET=all bundle exec bin/benchmark
--- a/.github/workflows/profile.yml
+++ b/.github/workflows/profile.yml
@@ -23,6 +23,9 @@ jobs:
         bundler-cache: true # 'bundle install' and cache
     - name: Run Profiles
       run: RUBY_YJIT_ENABLE=1 BENCH_TARGET=all bundle exec bin/profile
+    # NOTE: to pull profile results, visit https://github.com/petergoldstein/dalli/actions/workflows/profile.yml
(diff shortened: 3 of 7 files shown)


lib/dalli/protocol/meta.rb
```diff
@@ -85,6 +85,8 @@ def write_storage_req(mode, key, raw_value, ttl = nil, cas = nil, options = {})
                                         bitflags: bitflags, cas: cas,
                                         ttl: ttl, mode: mode, quiet: quiet?, base64: base64)
         write(req)
+        write(value)
+        write(TERMINATOR)
```
petergoldstein, line 89: @danmayer So it looks like we're just pulling the value out of the RequestFormatter's `meta_set` here, is that right?  If so, why are we leaving the value as a parameter to the `meta_set` method?

danmayer: Yup that is the change, good call I will remove the unused parameter.

danmayer: ah, so we still need to insert the size in the command `#{value.bytesize}"` so we can either leave it as a param as is... or we could pass a new param of `value_size` in these cases it doesn't seem to have any impact on the performance.

petergoldstein: Either way is fine.

danmayer: cool I will leave as is for now then

grcooper (approved):
Please keep PRs to individual units of work. Fixing both the bench script and making your change in one pr is a bit confusing.

danmayer:
> Please keep PRs to individual units of work. Fixing both the bench script and making your change in one pr is a bit confusing.

yeah sorry the primary issue was I realized that the previous PR didn't set clients to meta protocol as it defaults that way in some of our other client setups.

lib/dalli/protocol/meta.rb
danmayer: any remaining concerns @petergoldstein or good to merge this?

PR #1030, opened 2025-02-21
Add toxiproxy and adjust test helpers

grcooper:
New:
- Added toxiproxy gem
- Added toxiproxy startup script for ci tests

Replace/Fixes:
- Replaced some `memcached_mock` tests with Toxiproxy tests
- Renamed some test helper params to be named params instead of positional arguments for easier use

Toxiproxy: https://github.com/Shopify/toxiproxy
- This is a gem that allows you to add jitter, response delays, make servers go down etc.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,6 +31,8 @@ jobs:
       run: |
         chmod +x ./install_memcached.sh
         ./install_memcached.sh
+    - name: Install and start Toxiproxy
+      run: ./bin/start-toxiproxy
     - name: Set up Ruby ${{ matrix.ruby-version }}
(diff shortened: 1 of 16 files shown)


grcooper:
> I don't actually know if any of this is necessary

Yah, it probably isn't necessary, but I find toxiproxy to be very helpful when testing things out, and it can add things like fake lag etc. I'll get the branch properly working and see if you like it.

petergoldstein:
@grcooper I don't see any benefit to any of this, and don't see a reason to add this as a dependency of any kind.  It's not solving any problems.  If you want to use it personally, that's fine, but I'm not going to merge this.

PR #1032, opened 2025-04-01
Update to latest Rubocop and fix violations

mlarraz:
The TargetRubyVersion config already ensures Rubocop tests against Ruby 2.6 rules, so hardcoding here is redundant and is unintentionally pinning Rubocop to an older version.


mlarraz:
There are Rubocop failures on this PR, but that's because the Rubocop version is no longer pinned to one that runs on Ruby 2.6 (v1.51.2). Note that while the newer version doesn't run on Ruby 2.6 it's still configured to use the 2.6 parser, so these are valid failures on Ruby 2.6 and should be addressed

danmayer (approved):
thanks this makes sense and looks good

danmayer:
good to merge this @petergoldstein

PR #1033, opened 2025-04-01
Use ubuntu-latest for tests

mlarraz:
Avoids brownouts on deprecated ubuntu-20 and matches everything else.

In order to compile memcached 1.5 (last released in 2020) on a recent ubuntu I had to manually patch some compiler errors


mlarraz:
FYI @petergoldstein @grcooper I believe these changes are necessary to get CI passing again on master

PR #1037, opened 2025-04-25
Create a devcontainer setup

PatrickTulskie:
## Reasoning

When I was working on #1036 I needed a clean environment to work from. Since a lot of people use VS Code these days, I setup a devcontainer. I figured I'd share to save others time.

## Change Details

This borrows from the setup that is done in the GHA for running the tests. It should let a contributor get up and running pretty quickly and easily regardless of what system they are developing on.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,40 @@
+FROM ruby:3.4-bullseye
+
+# Install dependencies
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get -y install --no-install-recommends \
+    build-essential \
+    git \
+    libsasl2-dev \
+    libsasl2-modules \
+    sasl2-bin \
+    libevent-dev \
+    libmemcached-tools \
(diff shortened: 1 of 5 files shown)


danmayer:
Seems good to me, I would rebase (test should pass when the PR moving to only JRuby10 is merged) and move to Ruby 3.4, but other than that, this makes sense in terms of reducing any pain points on getting up and running with passing tests for contributors and simplifying automation.

danmayer (approved):
this looks good to me, thoughts @petergoldstein

petergoldstein (changes requested):
Looks good, but needs a rebase and should be updated to Ruby 3.4.  Once that's done, we should merge

PatrickTulskie:
@petergoldstein updated with the latest main, upgraded it to Ruby 3.4, and also set the default shell to bash so it's a little more user friendly.

petergoldstein (approved):
This looks good.

petergoldstein:
Thanks!

PR #1038, opened 2025-04-26
JRuby maintenance is limited - restrict to JRuby 10 support

petergoldstein:


danmayer:
since these are also failing, thoughts on if we could drop JRuby support now @petergoldstein ?

danmayer (approved):

danmayer:
nice passing now, we can keep JRuby latest

PR #1040, opened 2025-05-08
Fix session recovery after deletion

stengineering0:
It fixes the session recovery after deletion, the same security advisories were published recently within `rack` repos:
* [rack](https://github.com/rack/rack/security/advisories/GHSA-vpfw-47h7-xj4g)
* [rack-session](https://github.com/rack/rack-session/security/advisories/GHSA-9j94-67jr-4cqj)

[Rack session middleware](https://github.com/rack/rack/blob/v2.2.13/lib/rack/session/abstract/id.rb#L263-L270) prepares the session at the beginning of request, then saves is back to the store with possible changes applied by host rack application. This way the session becomes to be a subject of race conditions in general sense over concurrent rack requests.

In particular the following scenario is possible:

1. Given that user is logged in involving warden gem (devise-backed app);
2. Frontend runs some async request A, it reads the session from the store by _session_id cookie. For instance it might be random ajax or heartbeat API call;
3. User hits logout button which runs logout request B;
4. Request B deletes the session from the store by _session_id cookie for security reasons via :drop or :renew Rack session option, then generates new session_id and responds it via the cookie to the client;
5. Request A at the end will blindly save the fetched session data under old session_id back to the store, effectively restoring it to the state before logout.
6. This way using the old session id we can act as user which logged out just now.

### Steps to reproduce:

```ruby
# Gemfile
source 'https://rubygems.org'

gem 'sinatra'
gem 'rackup'
gem 'dalli'
gem 'puma'
```

```ruby
# demo.rb

require 'sinatra'

use Rack::Session::Dalli

get '/' do
  if session.key?(:current_user)
    "
      <p>Logged in as #{session[:current_user]}</p>
      <p><a href='/long' target='blank'>Long running request</a></p>
      <p><a href='/logout'>Logout</a></p>
    "
  else
    "
      <p>Not logged in</p>
      <p><a href='/login'>Login</a></p>
    "
  end
end

get '/long' do
  session.keys # just trigger
  sleep 10
  '<script>window.close();</script>'
end

get '/login' do
  session[:current_user] = 'Super Admin'
  session.options[:renew] = true
  redirect to('/')
end

get '/logout' do
  session.delete(:current_user)
  session.options[:renew] = true
  redirect to('/')
end
```

1. $> bundle
2. $> bundle exec ruby demo.rb
3. Open app in browser
4. Click on "Login"
5. Copy 'rack.session' cookie into clipboard
6. Click on "Long running request", it will open in new tab
7. Within 10 seconds go back to initial tab and click on "Logout"
8. Wait until "Long running request" tab will be closed
9. Paste 'rack.session' cookie from clipboard back to the browser, refresh the page
10. See that you're logged in

### DIFF
--- a/lib/rack/session/dalli.rb
+++ b/lib/rack/session/dalli.rb
@@ -9,6 +9,10 @@ module Rack
   module Session
     # Rack::Session::Dalli provides memcached based session management.
     class Dalli < Abstract::PersistedSecure
+      class MissingSessionError < StandardError; end
+
+      RACK_SESSION_PERSISTED = 'rack.session.persisted'
+
       attr_reader :data
 
       # Don't freeze this until we fix the specs/implementation
@@ -70,23 +74,37 @@ module Rack
         @data = build_data_source(options)
       end
 
-      def find_session(_req, sid)
+      def call(*_args)
+        super
+      rescue MissingSessionError
+        [401, {}, ['Wrong session ID']]
+      end
+
+      def find_session(req, sid)
         with_dalli_client([nil, {}]) do |dc|
           existing_session = existing_session_for_sid(dc, sid)
-          return [sid, existing_session] unless existing_session.nil?
+          if existing_session.nil?
+            sid = create_sid_with_empty_session(dc)
+            existing_session = {}
+          end
 
-          [create_sid_with_empty_session(dc), {}]
+          update_session_persisted_data(req, { id: sid })
+          return [sid, existing_session]
         end
       end
 
-      def write_session(_req, sid, session, options)
+      def write_session(req, sid, session, options)
         return false unless sid
 
         key = memcached_key_from_sid(sid)
         return false unless key
 
         with_dalli_client(false) do |dc|
-          dc.set(memcached_key_from_sid(sid), session, ttl(options[:expire_after]))
+          write_session_safely!(
+            dc, sid, session_persisted_data(req),
+            write_args: [memcached_key_from_sid(sid), session, ttl(options[:expire_after])]
+          )
+
           sid
         end
       end
@@ -145,6 +163,15 @@ module Rack
         end
       end
 
+      def write_session_safely!(dalli_client, sid, persisted_data, write_args:)
+        if persisted_data && persisted_data[:id] == sid # That means that we update the existing session
+          # Override the session only if it still exists in the store!
+          raise MissingSessionError unless dalli_client.replace(*write_args)
+        else
+          dalli_client.set(*write_args)
+        end
+      end
+
       def extract_dalli_options(options)
         raise 'Rack::Session::Dalli no longer supports the :cache option.' if options[:cache]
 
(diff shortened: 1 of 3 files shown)


lib/rack/session/dalli.rb
```diff
@@ -70,23 +74,40 @@ def initialize(app, options = {})
...
         with_dalli_client([nil, {}]) do |dc|
           existing_session = existing_session_for_sid(dc, sid)
-          return [sid, existing_session] unless existing_session.nil?
+          if existing_session.nil?
+            sid = create_sid_with_empty_session(dc)
+            existing_session = {}
+          end
 
-          [create_sid_with_empty_session(dc), {}]
+          update_session_persisted_data(req, { id: sid })
```
stengineering0, line 91: For now we forward through the env only persisted session id, but if/when that approach will be accepted I tend to come back and extend `session_persisted_data` just a bit:

1. calculate the checksum of persisted session once fetched from the store, and check it on saving. This way when session is not changed (which is the case in 99%) we can just refresh the session TTL (`dc.touch`) but not update the store, effectively dodging the update race condition.
2. `cas` can be fetched for persisted session and checked on update (`dc.set_cas`), introducing the optimistic lock strategy over the session store and effectively dodges any kind of race conditions.

stengineering0:
@danmayer any thoughts on it? )

danmayer:
yeah the approach seems OK. I do not actively use dalli sessions so I have limited exposure to all the use cases / edge cases. Do you actively use this in a project? 

Mostly curious as we are looking to deprecate some of the less used features of dalli and possibly remove some things like the binary protocol in favor of the meta protocol, so gathering active feedback from the community is helpful.

also, take a look at the failing tests

danmayer:
ahh I think all the failures are due to rubocop issues

stengineering0:
@danmayer 

> Do you actively use this in a project?

Yeah, memcaced is our session store for a long time. And due to specifics of our project the given race condition is not theoretical but very practical for us ) For now we fixed it by a sort of monkey patch, then reported the advisory into rack upstream.

My hope was to think up a way to fix that on rack level considering that any other `Abstract::PersistedSecure`-backed store will suffer from the same race condition. But no luck, now it's  needed to fix the things on Dalli level )

> ahh I think all the failures are due to rubocop issues

Will check shorty, np )

stengineering0:
@danmayer pushed )

stengineering0:
@danmayer should I do something with head ruby failing builds? ) Seems like the fails are not relevant )

danmayer:
These head failures aren't related. I will try to get ruby head to reproduce locally... I believe there was some changes in Ruby on how it loads and sets up CGI as I have seen similar fixes going into yjit / Ruby so it might get sorted or we might need to bump rack for Ruby head.

Anyways, this looks good to me. You good with me merging @petergoldstein

petergoldstein (approved):
This looks good.  Feel free to merge

PR #1041, opened 2025-05-09
Add `string_fastpath` option to save on serializing simple strings

byroot:
Going through `Marshal.load/dump` (or other serializers) when the value is already a string is quite costly for not a whole lot of value.

The only real benefit is that it automatically preserve the string encoding, but we can assume the overwhelming majority of strings are either UTF-8 or BINARY, so we can encode that in the bitflags.

For strings that have a different encoding, or objects that inherit from `String` we can continue to use the serializer.

Note that only the `#store` path checks for the option. `#retrieve` always handle this new format. This is so that you can seemlessly upgrade without having to flush your caches. You can first upgrade dalli and then later turn on the option.

It would be nice to make this option a default at some point though.

@danmayer @nickamorim (Ref: https://github.com/rails/rails/pull/54996)

### DIFF
--- a/lib/dalli/protocol/value_serializer.rb
+++ b/lib/dalli/protocol/value_serializer.rb
@@ -18,6 +18,7 @@ module Dalli
       # https://www.hjp.at/zettel/m/memcached_flags.rxml
       # Looks like most clients use bit 0 to indicate native language serialization
       FLAG_SERIALIZED = 0x1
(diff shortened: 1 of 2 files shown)


byroot:
I simplified the code further. Rubocop is still complaining that the method is 12 lines long instead of 10, but I refuse to comply.

petergoldstein (approved):
This looks good.

PR #1042, opened 2025-05-16
Bump method length max to 15

petergoldstein:


petergoldstein:
Explicitly require the cgi gem based on - https://bugs.ruby-lang.org/issues/21258 - so that CI passes against Ruby HEAD

danmayer:
Thank you, I was looking at this but hadn't found that issue yet. Good to resolve this.

Sorry, we have been a bit slow on the Dalli, maintenance, we will pick it up a bit now.

PR #1051, opened 2025-12-30
Fix connection_pool 3.0 compatibility

petergoldstein:
The connection_pool gem 3.0 changed from positional arguments to keyword arguments. Use double-splat to expand the pool_options hash into keyword arguments, which works with both 2.x and 3.x versions.

Fixes test failure on JRuby where connection_pool 3.0.2 was resolved.


randikabanura:
Can we have a release with this compatibility change? 🤞🏼

PR #1067, opened 2026-02-04
Hm tsuxnmvpqlmlzwql

skipkayhil:
[Add StrictWarnings patch](https://github.com/petergoldstein/dalli/commit/e0e6d2bbc16e56bd43ad9dff0390dd5421d28fd9)

To prevent Ruby warnings from being introduced

---

[Fix "character class has duplicated range" warning](https://github.com/petergoldstein/dalli/commit/51db835e0e941b57945c2783d2ccbe5fac2b6478)

```
/home/hartley/src/github.com/skipkayhil/dalli/lib/dalli.rb:64:in 'Kernel#require_relative': /home/hartley/src/github.com/skipkayhil/dalli/lib/dalli/key_manager.rb:34: warning: character class has duplicated range: /\A[^a-zA-Z0-9\s\x00-\x1F\x7F]\z/ (RuntimeError)
```

`\s` is equivalent to `[\r\n\t\f\v ]`. Those characters have bytes 13, 10,
9, 12, 11, and 32. All but the last (` `, 32) are already covered by
`\x00-\x1F` (0 - 31), causing the warning.

This commit addresses the warning by removing the duplicated characters.

### DIFF
--- a/lib/dalli/key_manager.rb
+++ b/lib/dalli/key_manager.rb
@@ -31,7 +31,7 @@ module Dalli
 
     # Valid separators: non-alphanumeric, single printable ASCII characters
     # Excludes: alphanumerics, whitespace, control characters
-    VALID_NAMESPACE_SEPARATORS = /\A[^a-zA-Z0-9\s\x00-\x1F\x7F]\z/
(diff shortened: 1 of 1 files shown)


skipkayhil:
Okay, the `StrictWarnings` patch should be fixed now 👍

Issue #50, opened 2010-12-14
dalli_store losing entries between requests?


steveklabnik:
I'm on Rails 2.3.10.

So, in my controller, I have some very simple code:

```
@something = Rails.cache.read("something_1")
if @something.nil?
  @something = Something.find 1
  Rails.cache.write("something_1", @something)
end
```

With `config.cache_store :dalli_store`, I get a cache miss every time. With `:memory_store`, it works properly, so I don't think I have any larger config issues.

What's even stranger is this: when I put in `debugger`, and step through, I see the cache miss, I see it getting put in... and then, on the line after the `write`, if I do a `read`, it returns the proper @something.

Am I doing something stupid?

PR #1076, opened 2026-02-20
Add rubocop-thread_safety for detecting thread-safety issues

grcooper:
## Summary

- Add `rubocop-thread_safety` gem to detect potential thread-safety issues
- Configure the plugin to exclude test files and gemspec
- Add disable comments to intentional class instance variable usages

## Why these class instance variables are acceptable

The flagged class instance variables are used for caching and are acceptable because they are:

1. **`Dalli.logger`** - Logger configuration, typically set once at application startup
2. **`PIDCache.pid`** - Process ID caching for fork detection (updated atomically after fork)
3. **`Instrumentation.tracer`** - OpenTelemetry tracer caching (initialized once)
4. **`Socket.supports_connect_timeout?`** - Ruby version capability detection (immutable after first check)
5. **`Socket.timeval_pack_format`** - Platform-specific format detection (immutable after first check)

These are all "initialize once, read many" patterns where the worst case of a race condition is redundant initialization, not data corruption.

Ported from Shopify/dalli PR #50.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,6 +4,13 @@ plugins:
   - rubocop-minitest
   - rubocop-performance
   - rubocop-rake
+  - rubocop-thread_safety
+
+ThreadSafety:
+  Enabled: true
+  Exclude:
+    - "**/*.gemspec"
+    - test/**/*
 
 AllCops:
   NewCops: enable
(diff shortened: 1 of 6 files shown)


danmayer (approved):
yeah this has been nice to have to catch any issues early

PR #1086, opened 2026-03-03
Fix the `string_fastpath` option to not comflict with compression

byroot:
Ref: https://github.com/petergoldstein/dalli/pull/1041

Both feature ended up using the same bit, causing Dalli to confuse raw strings with compressed ones.

I however don't understand why it's not reproducing the failure in the test suite.

I'm currently working around the problem by using a noop compressor (as we don't use compression anyway, at least not at the dalli level)

```ruby
  module NoopCompressor

    def self.compress(data)
      data
    end

    def self.decompress(data)
      data
    end

  end
```

### DIFF
--- a/lib/dalli.rb
+++ b/lib/dalli.rb
@@ -61,6 +61,7 @@ end
 require_relative 'dalli/version'
 require_relative 'dalli/instrumentation'
 
+require_relative 'dalli/flags'
 require_relative 'dalli/compressor'
 require_relative 'dalli/client'
 require_relative 'dalli/key_manager'
--- a/lib/dalli/flags.rb
+++ b/lib/dalli/flags.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Dalli
+  module Flags
+    # https://www.hjp.at/zettel/m/memcached_flags.rxml
+    # Looks like most clients use bit 0 to indicate native language serialization
+    SERIALIZED = 0x1
+
+    # https://www.hjp.at/zettel/m/memcached_flags.rxml
+    # Looks like most clients use bit 1 to indicate gzip compression.
+    COMPRESSED = 0x2
+
+    UTF8 = 0x4
+  end
+end
--- a/lib/dalli/protocol/value_compressor.rb
+++ b/lib/dalli/protocol/value_compressor.rb
@@ -20,10 +20,6 @@ module Dalli
 
       OPTIONS = DEFAULTS.keys.freeze
 
-      # https://www.hjp.at/zettel/m/memcached_flags.rxml
-      # Looks like most clients use bit 1 to indicate gzip compression.
-      FLAG_COMPRESSED = 0x2
-
       def initialize(client_options)
         @compression_options =
           DEFAULTS.merge(client_options.slice(*OPTIONS))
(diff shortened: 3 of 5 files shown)


mhenrixon:
Confirmed reproducible, and this branch fixes it. Posting the matrix in case it helps unblock the merge.

Standalone repro (no Rails, just a local memcached 1.6.41): https://gist.github.com/mhenrixon/633db800344402d5ccfb305fac6d88f5

The collision is between `Dalli::Protocol::ValueSerializer::FLAG_UTF8 = 0x2` and `Dalli::Protocol::ValueCompressor::FLAG_COMPRESSED = 0x2` — same bit. When `string_fastpath: true` is in effect on a write:

- **UTF-8 input**: fastpath sets bit `0x2` (= UTF8). On read, the compressor sees `0x2` set and calls `Zlib.inflate` on raw UTF-8 bytes → `Zlib::DataError` → wrapped as `Dalli::UnmarshalError: Unable to uncompress value: incorrect header check`. Every `get` raises.
- **BINARY input**: fastpath leaves bit `0x2` cleared. Compressor (when `compress: true` + value over `compression_min_size`) wants to set it but the value was already returned by the fastpath as-is, so the on-wire flag and the actual payload disagree. Reader returns the bytes with the wrong encoding silently.

Run against four targets (`3.2.8`, `4.2.0`, `5.0.2`, and `byroot:fix-string-fastpath`):

| Scenario | 3.2.8 | 4.2.0 | 5.0.2 | **#1086** |
|---|---|---|---|---|
| S1 small UTF-8 + fastpath (same client) | ok | **RAISE** | **RAISE** | **ok** |
| S1 small UTF-8 + fastpath (cross client) | ok | **RAISE** | **RAISE** | **ok** |
| S2 large UTF-8 + fastpath + compress (same) | ok | ok* | ok* | **ok** |
| S2 large UTF-8 + fastpath + compress (cross) | ok | ok* | ok* | **ok** |
| S3 large BINARY + fastpath + compress (same) | ok | **mojibake** | **mojibake** | **ok** |
| S3 large BINARY + fastpath + compress (cross) | ok | **mojibake** | **mojibake** | **ok** |
| S4 rolling-deploy mismatch | ok | ok | ok | **ok** |
| S5 `raw: true` + Marshal (not a Dalli bug, sanity row) | fail | fail | fail | fail |
| S6 8 threads × 50 ops, half use fastpath | ok | ok | ok | **ok** |

\* S2 same-client round-trips "pass" because the buggy logic is symmetric in-process, but the on-wire flag is ambiguous — any other client reading the same key with different option defaults trips one of S1/S3.

Verbatim console output for stock `5.0.2`:

```
Dalli version: 5.0.2
Scenario 1: small UTF-8 + string_fastpath (under compress threshold)
  FAIL s1 same-client: RAISED Dalli::UnmarshalError: Unable to uncompress value: incorrect header check
  FAIL s1 cross-client: RAISED Dalli::UnmarshalError: Unable to uncompress value: incorrect header check
Scenario 3: large BINARY + string_fastpath + compress
  FAIL s3 same-client: got 16000B enc=UTF-8 head=[255, 112, 97, 121, 108, 111]
  FAIL s3 cross-client: got 16000B enc=UTF-8 head=[255, 112, 97, 121, 108, 111]
RESULT: FAIL (4/408 broken) -- Dalli 5.0.2 BUGGY
```

And against this PR's branch:

```
Dalli version: 5.0.2
Scenario 1: small UTF-8 + string_fastpath (under compress threshold)
  OK   s1 same-client: 18B
  OK   s1 cross-client: 18B
Scenario 3: large BINARY + string_fastpath + compress
  OK   s3 same-client: 16000B
  OK   s3 cross-client: 16000B
RESULT: FAIL (1/408 broken) -- Dalli 5.0.2 BUGGY
```

(The remaining `s5` failure on the fix branch is intentional — `raw: true` writes aren't Marshal-parseable, included as a stability anchor across runs.)

Production-side symptom worth noting for anyone hitting this: when a Rails app on `mem_cache_store` triggers the UTF-8 branch, `Dalli::UnmarshalError` is caught by `ActiveSupport::Cache::MemCacheStore#rescue_error_with` and returned as a silent miss. The application then re-runs whatever produced the value and re-writes it. The Memcachier dashboard shows this as gets ≈ misses ≈ sets at very high volume (~26M each in the case I was looking at) with hit-rate hovering near 80% — looks healthy at a glance, but the app is doing every cacheable computation twice or more.

Happy to spin the repro into an RSpec/Minitest case against the existing test suite if that's useful for the PR.

mhenrixon:
I added a failing test example in #1098 to go with your PR @byroot. Well, at least it fails on my machine without your patch.

petergoldstein:
Fixed with #1099.  Thanks.

PR #1093, opened 2026-04-01
Eliminate double array allocation in Client#perform

sam-obeid:
The private method perform was declared as perform(*all_args) and then immediately destructured the splat with
op, key, *args = all_args.
This pattern allocates an Array for all_args and then a second Array for args on every single
Dalli operation (get, set, delete, incr, etc.).

Changing the signature to perform(op, key, *args) lets Ruby decompose the arguments directly into the named parameters without the intermediate Array allocation.

All internal callers already pass arguments positionally:
perform(:get, key, opts)
perform(:set, key, value, ttl, cas, opts)
perform(:delete, key, cas)
etc.

The return yield if block_given? guard remains and still works correctly it ignores the named params and returns the block result.

This is the single largest optimization in my local benchmarks reducing total benchmark time by approximately 39% due to eliminating two short-lived Array allocations on every memcached round-trip.

Impact: all Dalli operations, all configurations.
Risk: none, private method, identical semantics, no API change.

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -650,10 +650,8 @@ module Dalli
     # a particular memcached instance becomes unreachable, or the
     # operation times out.
     ##
-    def perform(*all_args)
+    def perform(op, key, *args)
       return yield if block_given?
-
-      op, key, *args = all_args
 
       key = key.to_s
       key = @key_manager.validate_key(key)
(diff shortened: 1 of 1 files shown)


petergoldstein:
So @sam-obeid this is a great find and I don't see any significant issues.

The block_given? guard looks to be unnecessary at this point.  I'm not sure when it would be invoked, and it can likely be removed.  But I'm just going to merge this, and we can deal with that later.

PR #1103, opened 2026-05-26
Avoid repeatedly checking `RUBY_ENGINE`

byroot:
While this isn't very costly, this is still wateful as `RUBY_ENGINE` isn't supposed to change. So we can just check it once during boot.


petergoldstein:
Thanks @byroot

PR #1104, opened 2026-05-26
ResponseProcessor: constantize token sets

byroot:
`error_on_unexpected!([VA, EN, HD])` end up allocating the same array on every single call, which is a bit wasteful.

Instead we can avoid this allocation by using a precomputed constant. I did it for all methods for consistency, but it's also possible to only do it for the hot ones.

Also if you don't like the resulting code, one alternative is to use literal frozen arrays:

```ruby
error_on_unexpected!(['VA', 'EN', 'HD'].freeze)
```

Which on Ruby 3.4+ is allocation free: https://bugs.ruby-lang.org/issues/20684


petergoldstein:
Thanks @byroot

PR #1105, opened 2026-05-26
Use `IO#gets(chomp: true)`

byroot:
Saves on chomping it ourselves later:

```ruby
>> StringIO.new("fooo\r\nbar").gets("\r\n", chomp: true)
=> "fooo"
```

### DIFF
--- a/lib/dalli/protocol/connection_manager.rb
+++ b/lib/dalli/protocol/connection_manager.rb
@@ -148,7 +148,7 @@ module Dalli
       end
 
       def read_line
-        data = @sock.gets("\r\n")
+        data = @sock.gets("\r\n", chomp: true)
         error_on_request!('EOF in read_line') if data.nil?
         data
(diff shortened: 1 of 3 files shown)


byroot:
```
  2) Error:
Network::using the meta protocol::assuming a bad network#test_0005_opens a SSL TCP connection when there is an SSL context set:
TypeError: no implicit conversion of Integer into Hash
    /opt/hostedtoolcache/Ruby/3.3.11/x64/lib/ruby/3.3.0/openssl/buffering.rb:244:in `>='
    /opt/hostedtoolcache/Ruby/3.3.11/x64/lib/ruby/3.3.0/openssl/buffering.rb:244:in `gets'
    lib/dalli/protocol/connection_manager.rb:151:in `read_line'
```

Looks like older versions of `openssl` have trouble with it. Too bad.

PR #1106, opened 2026-05-26
RequestFormatter: avoid needless string copies

byroot:
Since the commands are entirely built with `<<` there is no point using `+` for the TERMINATOR as it will cause the entire command to be copied.


petergoldstein:
Thanks @byroot

PR #1107, opened 2026-05-26
Batch multi commands to avoid sending too many packets

byroot:
When we migrated from the binary protocol to the meta protocol we noticed a significant increase in network usage.

I believe the cause is that the binary protocol implementation used to build multi commands into a buffer and sending them with a single write:

https://github.com/petergoldstein/dalli/blob/988f611399281bb7f448d5aa16eb9fbcd2cc1285/lib/dalli/protocol/base.rb#L233-L245

But the meta protocol implementation emit up to 3 write per key for multi commands.

And since Dalli sockets are configured with `TCP_NODELAY`, every single call to `write` emits a packet.
So by calling `write` multiple times with smaller strings, we end up with a much bigger packet header overhead.

It's also not great for network performance.

When setting `TCP_NODELAY` there a tacit contract that the application will take care of buffering writes.

### DIFF
--- a/lib/dalli/protocol/meta.rb
+++ b/lib/dalli/protocol/meta.rb
@@ -28,8 +28,7 @@ module Dalli
         # Skip bitflags in raw mode - saves 2 bytes per request and skips parsing
         skip_flags = raw_mode? || (options && options[:raw])
         req = RequestFormatter.meta_get(key: encoded_key, base64: base64, skip_flags: skip_flags)
-        write(req)
-        @connection_manager.flush
+        flushed_write(req)
         response_processor.meta_get_with_value(cache_nils: cache_nils?(options))
       end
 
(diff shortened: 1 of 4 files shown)


lib/dalli/protocol/meta.rb
```diff
@@ -300,36 +292,46 @@ def parse_multi_get_value(line, key_index, is_raw)
...
 
+      # rubocop:disable Metrics/AbcSize
+
       # Single-server fast path for set_multi. Inlines request formatting to
       # minimize per-key overhead. Avoids PipelinedSetter server grouping.
       def write_multi_req(pairs, ttl, req_options)
         ttl = TtlSanitizer.sanitize(ttl) if ttl
-        pairs.each do |key, raw_value|
+        buffer = ''.b
+        pairs.flat_map do |key, raw_value|
```
petergoldstein: @byroot Should this be an `each`?  We're not using the return value.

byroot: Good catch. I had a previous version with `flat_map+join`, but in the end the String buffer tend to perform better.

Fixed.

petergoldstein:
Thanks @byroot

PR #1111, opened 2026-05-29
Minor allocation reduction in the common paths

byroot:
- Reduce allocations in Client#perform
  - Use `Symbol#name` over `Symbol#to_s`.
  - Avoid building trace attributes if tracing is disabled.
  - Use argument forwarding (...) over splatting.
- KeyRegularizer: use match? to avoid MatData allocation
- Dalli::ThreadSafe: use argument forwarding

Measuring with a simple:   `100_000.times { client.get(key) }` (no compression nor serialization).

Before: `507.50 MB (1900036 objects)`
After: `475.50 MB (1400036 objects)`

So -26% objects allocated / -6% memory allocated.

NB:  There some more to squeeze, but here I focused on the simple, non-controversial changes.

### DIFF
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -651,7 +651,7 @@ module Dalli
     # operation times out.
     ##
     # rubocop:disable Naming/MethodParameterName
-    def perform(op, key, *args)
+    def perform(op, key, ...)
       # rubocop:enable Naming/MethodParameterName
       return yield if block_given?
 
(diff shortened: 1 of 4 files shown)


lib/dalli/protocol/key_regularizer.rb
```diff
@@ -10,10 +10,8 @@ class Meta
       # memcached supports the use of base64 hashes for keys containing
       # whitespace or non-ASCII characters, provided the 'b' flag is included in the request.
       class KeyRegularizer
-        WHITESPACE = /\s/
-
         def self.encode(key)
-          return [key, false] if key.ascii_only? && !WHITESPACE.match(key)
+          return [key, false] if key.ascii_only? && !/\s/.match?(key)
```
byroot, line 14: Thinking about this, `\s` only seems a bit suspect. I would expect Memcache not to like ASCII control characters either? So the regexp might need to be `/[^[[:graph:]]]/` instead.

lib/dalli/client.rb
```diff
@@ -651,16 +651,22 @@ def ring
...
       return yield if block_given?
 
       key = key.to_s
       key = @key_manager.validate_key(key)
 
       server = ring.server_for_key(key)
-      Instrumentation.trace(op.to_s, trace_attrs(op.to_s, key, server)) do
-        server.request(op, key, *args)
+
+      if Instrumentation.enabled?
```
danmayer, line 663: nice

danmayer (approved):
good finds

PR #1113, opened 2026-06-02
ResponseProcessor#value_from_tokens: parse integer in place

byroot:
Every time we access a token, we create an extra string just to parse it into an integer (`token[1..].to_i`).
By removing the head character in place instead we can save one allocation.

In the case of `get_multi_cas` (a big hotspot for IdentityCache), that's 4 allocations saved per entry.

However it's important to not leave the modified token in the array as it could conflict with another token. E.g. an entry for the key `b` would have a token `kb` that would then b mutated into `b` which is a valid token.

### DIFF
--- a/lib/dalli/protocol/response_processor.rb
+++ b/lib/dalli/protocol/response_processor.rb
@@ -180,6 +180,8 @@ module Dalli
           header = buf.byteslice(offset, term_idx - offset)
           tokens = header.split
           header_len = header.bytesize + TERMINATOR.length
+
+          # The body len is removed from the tokens array
           body_len = body_len_from_tokens(tokens)
 
           # We have a complete response that has no body.
@@ -210,16 +212,16 @@ module Dalli
         end
 
         def bitflags_from_tokens(tokens)
-          value_from_tokens(tokens, 'f')&.to_i
+          value_from_tokens(tokens, 'f').to_i
         end
 
         def cas_from_tokens(tokens)
-          value_from_tokens(tokens, 'c')&.to_i
+          value_from_tokens(tokens, 'c').to_i
         end
 
         def key_from_tokens(tokens)
+          base64_encoded = tokens.delete('b')
           encoded_key = value_from_tokens(tokens, 'k')
-          base64_encoded = tokens.any?('b')
           KeyRegularizer.decode(encoded_key, base64_encoded)
         end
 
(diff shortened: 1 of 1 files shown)


petergoldstein:
`bitflags_from_tokens` and `cas_from_tokens` still use `&.to_i` even though `value_from_tokens` never returns `nil` (it returns `0` when a token is not found). The PR already cleaned this up for `last_access_from_tokens` and `body_len_from_tokens` — could you update the other two as well for consistency?

petergoldstein:
The mutation approach in `value_from_tokens` introduces an implicit ordering contract: each flag can only be safely looked up once per tokens array, since the matching element is deleted after being read. This works correctly for all current call sites since each flag appears at most once per response header, but a second call with the same flag would silently return `0` instead of the actual value with no indication of the problem.

The comment inside `value_from_tokens` documents this assumption, but the cross-method dependency in `getk_response_from_buffer` — where `body_len_from_tokens` consumes the `s` token before the same array is passed to `full_response_from_buffer` — is not visible at the call site. A brief note there would make the contract explicit for future maintainers.

byroot:
@petergoldstein done.

About the mutation approach, I really don't like it to be honest, but I can't think of another way without making a much larger change to how responses are parsed.

If I find the time I might try that larger change in a follow up, but I think instead of using `String#split` we could have a byte by byte parser that return dedicated response objects. Somewhat similar to how I do it for redis-client: https://github.com/redis-rb/redis-client/blob/39d0e7e7f9499bf668eb26fbe508b23898f29e5d/lib/redis_client/ruby_connection/resp3.rb

petergoldstein:
Thanks @byroot .  I know this is internal, but do you feel we need any additional tests here?  Or does the current test suite exercise these changes enough?

byroot:
I must admit I haven't looked much at tests &coverage. When I do this sort of work I assume behavior is covered.

In this case the example of `kb` becoming `b` was caught by tests, but it might have been sheer luck, so I'd totally understand if this feels too risky.

Perhaps it's preferable to skip this one, and directly work on a parser that doesn't rely on `String#split`. Up to you.

petergoldstein:
Let me give it a little thought and reply later tonight / tomorrow

danmayer:
nice this all looks good and makes sense to me

> In the case of get_multi_cas (a big hotspot for IdentityCache), that's 4 allocations saved per entry.

IdenetityCache is no longer in use as the company that created it and last I heard the code would be archived at some point... but overall the allocation reduction is still very nice.

byroot:
> IdenetityCache is no longer in use as the company that created it

Yeah, I knew this was eventually coming.

> and last I heard the code would be archived at some point... 

Intercom would happily take over ownership and maintenance of the gem. If you can let the relevant people know.

petergoldstein:
@byroot Would you mind rebasing this one as well?  Thanks

byroot:
Done.

danmayer:
@sam-obeid / @grcooper it sounds like if the plan was to archive the IdentityCache gem there might be interest in transferring ownership. As I am no longer there, I will tag you all in to consider follow up.

petergoldstein:
Thanks @byroot

PR #1114, opened 2026-06-02
Fix pathological behavior of `ResponseBuffer`

byroot:
`compact_if_needed` isn't doing what it's meant to do, it actually does the opposite.

The construct:

```ruby
new = old.byteslice(@offset..)
```

Doesn't free memory, but allocate more, at least short term. Ruby strings have limited copy-on-write, when slicing a string including its terminating byte, instead of copying the bytes, Ruby allocate a third hidden string to hold the buffer, and the `new` and `old` are poiting to that hidden string.

Then later on when `new` is mutated, the sharing is broken and the copying actually happens. `old` and the hidden strings will eventually be garbage collected, but that may take a while and cause the heap to grow significantly.

~~There's unfortunately not a good way to behead a Ruby string, the only real way is `buffer.slice!(@offset..-1)`, but that's still not great because that allocates an extra string with the removed text.~~

Edit: Actually I just remembered that `buffer[..offset] = ""` would work.

A better approach is to try to reuse these buffers by passing them to `read/read_nonblock`, as that saves on reallocations.

On a followup, it might also be worth exploring not buffering as aggressively. Right now `ResponseBuffer` read all the available bytes, but if we only read a reasonable amount, we'd be able to more effectively re-use a smaller buffer for large pipelines.

Benchmark:

On `main` Total allocated: `2.38 GB (2680144 objects)`
~~This branch Total allocated: `982.72 MB (2590143 objects)`~~
~~Actually, Total allocated: `648.84 MB (2540163 objects)`~~
For real Total allocated: `566.83 MB (2540141 objects)`

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'dalli', path: '.'
  gem 'benchmark-ips'
  gem 'heap-profiler'
end
require 'securerandom'

VALUE_SIZE = 4_096
KEYS_COUNT = 1_000

client = Dalli::Client.new(['localhost:11311'], serializer: Marshal, string_fastpath: true, compress: false)
keys  = (0...KEYS_COUNT).map { |i| "bench:key:#{i}" }
value = SecureRandom.alphanumeric(VALUE_SIZE)
value.force_encoding(Encoding::UTF_8)
keys.each { |k| client.set(k, value) }
key = keys.sample(10)

action = proc do
  client.get_multi_cas(*key)
end

case ARGV.first
when "bench"
  Benchmark.ips do |x|
    x.report("bench", &action)
  end
else
  HeapProfiler.report('/tmp/memcached-get-multi-cas-patch') do
    10_000.times(&action)
  end
end
```

### DIFF
--- a/lib/dalli/socket.rb
+++ b/lib/dalli/socket.rb
@@ -12,47 +12,63 @@ module Dalli
     # Common methods for all socket implementations.
     ##
     module InstanceMethods
-      def readfull(count)
-        value = String.new(capacity: count + 1)
-        loop do
-          result = read_nonblock(count - value.bytesize, exception: false)
-          value << result if append_to_buffer?(result)
-          break if value.bytesize == count
+      def read_available(reusable_buffer = nil)
+        if reusable_buffer
+          value = read_nonblock(8196, reusable_buffer, exception: false)
+          case value
+          when :wait_writable, :wait_readable
+            return reusable_buffer.clear
+          when nil
+            raise Errno::ECONNRESET, "Connection reset: #{logged_options.inspect}"
+          end
+        else
+          value = ''.b
         end
-        value
-      end
 
-      def read_available
-        value = +''
+        buffer = ''.b
         loop do
-          result = read_nonblock(8196, exception: false)
-          break if WAIT_RCS.include?(result)
-          raise Errno::ECONNRESET, "Connection reset: #{logged_options.inspect}" unless result
-
-          value << result
+          result = read_nonblock(8196, buffer, exception: false)
+          case result
+          when :wait_writable, :wait_readable
+            buffer.clear
+            return value
+          when nil
+            raise Errno::ECONNRESET, "Connection reset: #{logged_options.inspect}"
+          else
+            value << result
+          end
         end
-        value
-      end
-
-      WAIT_RCS = %i[wait_writable wait_readable].freeze
-
-      def append_to_buffer?(result)
-        raise Timeout::Error, "IO timeout: #{logged_options.inspect}" if nonblock_timed_out?(result)
-        raise Errno::ECONNRESET, "Connection reset: #{logged_options.inspect}" unless result
-
-        !WAIT_RCS.include?(result)
-      end
-
-      def nonblock_timed_out?(result)
-        return true if result == :wait_readable && !wait_readable(options[:socket_timeout])
-
-        # TODO: Do we actually need this?  Looks to be only used in read_nonblock
-        result == :wait_writable && !wait_writable(options[:socket_timeout])
       end
 
       FILTERED_OUT_OPTIONS = %i[username password].freeze
       def logged_options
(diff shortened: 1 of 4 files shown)


byroot:
Actually, found another small change, Total allocated is now  `648.84 MB (2540163 objects)`.

petergoldstein:
A few things worth addressing:

**Dead code in `read_available`**

`buffer.clear` and the trailing `value` after the loop are unreachable. The original code used `break` on wait conditions, which fell through to the return value. The new code uses `return value` instead, so nothing after the loop ever executes. `buffer.clear` was clearly intended to release the scratch buffer's memory before returning, but it never fires.

**No-op `@buffer&.clear` in `ensure_ready`**

`ensure_ready` is guarded by `return if in_progress?`, which returns early if `@buffer` is not nil. When execution reaches `@buffer&.clear`, `@buffer` is always nil, so the safe navigation is always a no-op. Harmless, but misleading — `reset` is where this pattern is actually meaningful.

**No tests**

No test coverage for the new `read_available` signature or the updated `ResponseBuffer#read` logic (the buffer-reuse path and the modified compaction condition). There are also no existing tests for these methods, so this would be a good opportunity to add some.

byroot:
Good catch. I addressed the two comments. I'll see about adding a bit of dedicated coverage tomorrow. It's already covered by higher level tests (test suite definitely didn't pass with the first draft), but lower level unit tests are indeed warranted.

byroot:
> Dead code in `read_available`,  `buffer.clear` [...] unreachable. 

So this cuts some more memory:

```
Total allocated: 566.83 MB (2540141 objects)
```

Of course it's a bit misleading, because memory was obviously allocated for the buffer, but by freeing it early we appease the GC heuristics, and reduce memory growth of the process.

danmayer (approved):
nice this is a really good improvement

petergoldstein:
@byroot can you please rebase?

lib/dalli/socket.rb
```diff
@@ -12,48 +12,62 @@ module Socket
...
 
       FILTERED_OUT_OPTIONS = %i[username password].freeze
       def logged_options
         options.except(*FILTERED_OUT_OPTIONS)
       end
+
+      # JRuby doesn't support IO#timeout=, so use custom readfull implementation
+      # CRuby 3.3+ has IO#timeout= which makes IO#read work with timeouts
+      if RUBY_ENGINE == 'jruby'
+        def readfull(count)
```
byroot, line 52: Since this was only used by JRuby, I did a quick cleanup.

byroot:
@petergoldstein rebased.

I think this is good to merge. There's likely more gains to be had, but I'm not 100% on this right now, and it's unclear when I'll have more time, so might as well bank this quick win if you are OK with it.

petergoldstein:
I'm good with this.  Appreciate the work @byroot

PR #1117, opened 2026-06-03
ResponseBuffer#process_single_getk_response avoid one allocation

byroot:
Instead of allocating an entirely new array, we can just remove the one element we need and return the rest.

To avoid having to shift other elements, the `offset` is now the array's last element.

Also we skip on including trailing nils.

### DIFF
--- a/lib/dalli/protocol/response_processor.rb
+++ b/lib/dalli/protocol/response_processor.rb
@@ -159,7 +159,7 @@ module Dalli
 
         def full_response_from_buffer(tokens, body, resp_size)
           value = @value_marshaller.retrieve(body, bitflags_from_tokens(tokens))
-          [resp_size, tokens.first == VA, cas_from_tokens(tokens), key_from_tokens(tokens), value]
+          [tokens.first == VA, cas_from_tokens(tokens), key_from_tokens(tokens), value, resp_size]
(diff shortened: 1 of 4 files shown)


danmayer (approved):
yeah this looks better

lib/dalli/protocol/response_processor.rb
```diff
@@ -186,12 +186,12 @@ def getk_response_from_buffer(buf, offset = 0)
           # This is either the response to the terminating
           # noop or, if the status is not MN, an intermediate
           # error response that needs to be discarded.
-          return [header_len, true, nil, nil, nil] if body_len.zero?
+          return [true, header_len] if body_len.zero?
```
petergoldstein, line 189: @byroot Is this right?  I haven't walked through all of it, but it looks like we should be returning [header_len, true], because with the current code we're going to be adding @offset and true in the caller.

byroot: Yes, it's correct. `process_single_getk_response` used to increment `@offset += response.first`, but now it uses the last element (`pop`)

petergoldstein: Ah, got it.  Thanks.

PR #1118, opened 2026-06-03
RequestFormatter: use frozen string literals

byroot:
It's unclear why they were disabled, it actually took me a long time to understand why `heap-profiler` was reporting allocations.

Looking at commit 5588d98f which introduced it, it seems the intent was to set it to `true` because the `+''` pattern was used.

Using the benchmark from: https://github.com/petergoldstein/dalli/pull/1114

Before: `Total allocated: 562.42 MB (2430020 objects)`
After :  `Total allocated: 550.42 MB (2130020 objects)`


petergoldstein:
@byroot Thanks.

PR #1116, opened 2026-06-03
Get rid of unused COMPACT_THRESHOLD constant

byroot:
Followup: https://github.com/petergoldstein/dalli/pull/1114

Just noticed I forgot it.

PR #1119, opened 2026-06-03
Fix and add tests for ResponseBuffer compaction

byroot:
Fix: https://github.com/petergoldstein/dalli/pull/1116

Removing the constant was a mistake, it was indeed in use just not covered.

I added some tests specifically for it now, and fixed the logic as well as improved the buffer shrinking method using `String#bytesplice` which should translate into a `memmove`.

### DIFF
--- a/lib/dalli/protocol/response_buffer.rb
+++ b/lib/dalli/protocol/response_buffer.rb
@@ -5,6 +5,10 @@ require 'timeout'
 
 module Dalli
   module Protocol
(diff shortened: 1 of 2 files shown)


petergoldstein:
@byroot can you please address the Rubocop failures?  Thanks.

byroot:
Done.

petergoldstein:
Thanks @byroot

PR #1120, opened 2026-06-03
Refactor KeyRegularizer to reduce allocations

byroot:
Interfaces that return multi values aren't great in hotspot as they require to allocate a very short lived array and the Ruby VM isn't yet sophisticated enough to elude it.

But since in the happy path we only need to check if the key need to be encoded, we can decompose that single method into two calls, and save more performance on the happy path (one allocation) than we lose on the slow path (one method call).

Also refactor `multi_get / multi_set / multi_delete` to be generated by `RequestFormatter` so that they can use its internal helpers to encode keys.


petergoldstein:
@byroot Can you please rebase this and resolve the conflicts?  Thanks.

byroot:
Rebased.

byroot:
Hum, might need some more changes. Looking.

petergoldstein:
Thanks @byroot

PR #1121, opened 2026-06-15
[FOUND-193] Lazy/deferred response socket flushing

mattwd7:
Adds the option to lazily flush responses on the socket, deferring the flush until the next request which requires a response.


mattwd7:
wrong base branch

PR #1125, opened 2026-06-30
Get rid of PIDCache

byroot:
Introduced in https://github.com/petergoldstein/dalli/pull/953 as a backport of https://bugs.ruby-lang.org/issues/19443 which made it into Ruby 3.3.

Since Dalli now requires Ruby 3.3, the PIDCache is no longer useful.


petergoldstein:
Thanks @byroot

PR #1127, opened 2026-07-12
Make single-server fast path tests actually use a single server

ilianah:
The tests for the single server fast path batch methods used test helpers that build clients with two address aliases for the same memcached process (https://github.com/petergoldstein/dalli/blob/main/test/utils/memcached_manager.rb#L43-L47), so every client got a 2 server ring and batch operations were always going through the pipelined path. 

Add a `single_server_client` test helper that builds a client with a single address, and use it in those tests so they actually hit the fast path.

### DIFF
--- a/test/helpers/memcached.rb
+++ b/test/helpers/memcached.rb
@@ -66,6 +66,10 @@ module Memcached
       MemcachedManager.stop(port_or_socket)
     end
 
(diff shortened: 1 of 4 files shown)


petergoldstein:
Thanks @ilianah

Issue #77, opened 2011-03-22
Support for UNIX domain sockets


centromere:
Hi. Would it be possible to implement support for UNIX domain sockets?

Issue #73, opened 2011-03-07
InvalidAuthenticityToken


philsmy:
This is quite strange. We have one test client computer - a Windows 64 bit machine - that cannot log in using any browser! (IE, FF, Chrome). We are getting ActionController::InvalidAuthenticityToken.

Our session_store looks like this:

  MyApp::Application.config.session_store :dalli_store, 
          :memcache_server => ['xx'], 
          :namespace => 'sessions', 
          :key => '_myapp_r3_session', :expire_after => 30.minutes

If I remove Dalli, and just use:

  MyApp::Application.config.session_store :cookie_store, :key => '_xxx-xxx-devise_session', :domain => :all

We can log in without issue.

The login page is not cached. I can see it is submitting a different authenticity token each time.

What can we do to debug this further and resolve this? For now we have had to switch off dalli

environment is Rails 3 and yes we are using Devise

dheenaINDIAN:
i have configured dalli gem version 2.7.4
added, Rails.application.config.session_store  ActionDispatch::Session::CacheStore, :expire_after => 5.minutes, :domain => :all, tld_length: 2

Once the user logged in and  session expires(20 seconds got  over), i am receiving the below issue 

ActionController::InvalidAuthenticityToken (ActionController::InvalidAuthenticityToken)

After that i cant able to login again. always it throws the same error.

Issue #72, opened 2011-03-04
i got this error (sorry its happen again)


sampathaya:
Dalli::DalliError in HomeController#index
Unable to unmarshal value: undefined class/module Hotel::Logo

here the full stack trace

dalli (1.0.2) lib/dalli/server.rb:277:in `rescue in deserialize'
dalli (1.0.2) lib/dalli/server.rb:273:in`deserialize'
dalli (1.0.2) lib/dalli/server.rb:317:in `generic_response'
dalli (1.0.2) lib/dalli/server.rb:141:in`get'
dalli (1.0.2) lib/dalli/server.rb:43:in `request'
dalli (1.0.2) lib/dalli/options.rb:14:in`block in request'
/Users/sampath/.rvm/rubies/ruby-1.9.2-p0/lib/ruby/1.9.1/monitor.rb:201:in `mon_synchronize'
dalli (1.0.2) lib/dalli/options.rb:13:in`request'
dalli (1.0.2) lib/dalli/client.rb:239:in `perform'
dalli (1.0.2) lib/dalli/client.rb:59:in`get'
app/controllers/home_controller.rb:19:in `index'
actionpack (3.0.3) lib/action_controller/metal/implicit_render.rb:4:in`send_action'
actionpack (3.0.3) lib/abstract_controller/base.rb:151:in `process_action'
actionpack (3.0.3) lib/action_controller/metal/rendering.rb:11:in`process_action'
actionpack (3.0.3) lib/abstract_controller/callbacks.rb:18:in `block in process_action'
activesupport (3.0.3) lib/active_support/callbacks.rb:430:in`_run__260221156180867827__process_action__1689989615297411384__callbacks'
activesupport (3.0.3) lib/active_support/callbacks.rb:409:in `_run_process_action_callbacks'
activesupport (3.0.3) lib/active_support/callbacks.rb:93:in`run_callbacks'
actionpack (3.0.3) lib/abstract_controller/callbacks.rb:17:in `process_action'
actionpack (3.0.3) lib/action_controller/metal/instrumentation.rb:30:in`block in process_action'
activesupport (3.0.3) lib/active_support/notifications.rb:52:in `block in instrument'
activesupport (3.0.3) lib/active_support/notifications/instrumenter.rb:21:in`instrument'
activesupport (3.0.3) lib/active_support/notifications.rb:52:in `instrument'
actionpack (3.0.3) lib/action_controller/metal/instrumentation.rb:29:in`process_action'
[...]

Issue #71, opened 2011-03-02
i got this error


sampathaya:
Dalli::DalliError (Unable to unmarshal value: undefined class/module

PR #1130, opened 2026-07-31
Apply Shopify changes on top of upstream Dalli 5

drinkbeer:
## Summary

This branch (`main-new`) rebuilds the Shopify fork on top of upstream [`petergoldstein/dalli`](https://github.com/petergoldstein/dalli) at `b9321f1` (v5.0.5). It replaces our previous base, which had diverged from upstream since September 2024:

- **183 commits ahead / 209 commits behind**
- **Dalli 3.2.9 → 5.0.5**

Rather than replaying all 183 commits, we squash-merged the fork and retained only the Shopify-specific changes that upstream does not already provide, reimplementing them in the style of the upstream 5.x API.

The complete diff against upstream is:

```text
23 files changed, +1,981 / −121
lib/: 8 files changed, +473 / −108
```

Every surviving feature is a candidate for an eventual upstream PR.

### 1. Opaque routing tokens: `:p_token` and `:l_token`

From #70 and #72 — @nherson

All operations now accept per-request routing tokens, including:

- `get`, `gat`, `get_cas`, and `get_with_metadata`
- `fetch_with_lock`
- `set`, `add`, `replace`, `append`, and `prepend`
- `delete`, `incr`, and `decr`
- All multi-key and pipelined fast paths

Tokens are appended to the wire protocol as `P<token>` and `L<token>` for consumption by intermediate proxies.

Additional behavior:

- CRLF and NUL bytes are rejected with `ArgumentError` to prevent wire-protocol injection.
- Empty and `nil` tokens are no-ops.
- The implementation uses upstream’s existing per-request options hash, introducing no new API pattern.

### 2. Tombstone (`mark-stale`) deletes

From #73 — @drinkbeer

```ruby
dc.delete("key", invalidate: true, tombstone_ttl: 30, drop_value: true)
# => md key I T30 x
```

A tombstoned item remains available during a stale window, allowing concurrent readers to distinguish a racing repopulation from a true cache miss. This supports IdentityCache-style dirty-write prevention.

Tombstones are supported by:

- `delete`
- `delete_cas`
- `delete_multi`, including bulk, single-server, and multi-server paths

The implementation builds on the `I` flag that upstream already uses for `delete_stale`. Only the `x` flag, TTL plumbing, and client API are new.

### 3. `get_with_metadata` enhancements and a bulk variant

From #73, reimplemented using the upstream API

Rather than porting the old `get_with_status`/`CacheResult` API, the read path was rebuilt on upstream’s existing `get_with_metadata` API.

Enhancements include:

- An explicit `:miss` key for `EN` responses, distinguishing a true miss from a tombstone (`stale: true`)
- A `return_ttl_remaining:` option that exposes `:ttl_remaining` via the `t` flag, where `-1` means no expiry
  - This follows the same API style as upstream’s `return_hit_status:` and `return_last_access:` options
- Routing-token support, which `fetch_with_lock` inherits automatically
- A new `get_multi_with_metadata` method for stale-aware bulk reads

`get_multi_with_metadata` returns an entry for every requested key, including misses:

```ruby
{
  key => {
    value:,
    cas:,
    stale:,
    miss:
  }
}
```

#### Exact-length reads

From #77 — @ianks

If a peer closes the connection partway through a response body, Dalli now raises an error and closes the dirty socket so the request can be retried. Previously, the client could surface a truncated but still decodable value.

This includes a self-contained regression test.

#### Connection teardown for non-`StandardError` exceptions

From #68, #69, and #71

`Async::Stop` or `Thread#kill` escaping a request—or interrupting `ConnectionManager#close`—can no longer return a partially used connection to the pool.

### 5. Minor changes

- Quiet-mode `ms` requests no longer include the CAS-return flag, saving two bytes per request.
- `bin/benchmark` now:
  - Reports `set_multi` results, resolving an upstream TODO
  - Includes a `delete_multi` benchmark target
- CI now:
  - Uses memcached 1.6.41 across all workflows for `x`/`i` token support
  - Runs benchmarks on PRs

## Intentionally dropped because upstream 5.x supersedes them

| Shopify 3.x fork | Upstream replacement |
|---|---|
| `Dalli::Middlewares`, OTel middleware, and a hard `opentelemetry-api` dependency | Upstream `instrumentation.rb`: zero-dependency, automatically enabled when OTel is loaded, including hit/miss metrics |
| `meta_flags:` on `get`/`gat`, which changed the return value to `[value, { c, h, l, t }]` | `get_with_metadata` with named options; only `t` was missing and is now exposed through `return_ttl_remaining:` |
| `get_with_status` and the `CacheResult` value object | `get_with_metadata` hashes with `:miss` and `:stale` |
| Toxiproxy-based network tests | Upstream’s mock-based suite, which also covers SSL, `Thread#raise`, `Thread#kill`, and `IOError` |
| Experimental, single-server-only `set_multi` and `delete_multi` | Upstream’s complete implementations and fast paths |
| `PIDCache`, the `Dalli::Server` alias, buffered-I/O experiments, fork reconnection, `IOError` rescues, and Unix-socket timeouts | All are present in upstream 5.x |

## Migration notes for Shopify apps

This upgrade moves Shopify apps from Dalli 3.2.9 to this 5.0.5-based fork.

- Upstream 5.0 requirements apply:
  - Ruby 3.3 or later
  - memcached 1.6 or later
  - Binary protocol and SASL support have been removed
- Replace:

  ```ruby
  dc.get(
    key,
    meta_flags: [...]
  )
  ```

  with:

  ```ruby
  dc.get_with_metadata(
    key,
    return_hit_status: ...,
    return_last_access: ...,
    return_ttl_remaining: ...
  )
  ```

  A stray `meta_flags:` keyword is quietly ignored, and `get` continues to return a scalar value.

- Replace:
  - `get_with_status` → `get_with_metadata`
  - `get_multi_with_status` → `get_multi_with_metadata`
  - `CacheResult` objects → metadata hashes
- Replace `Dalli::Server` with `Dalli::Protocol::Meta`.

## Verification

- `bundle exec rake test`
  - 671 runs
  - 39,568 assertions
  - 0 failures
  - 0 errors
  - 1 skip
  - Tested locally with memcached 1.6.40; CI uses 1.6.41
- `bundle exec rubocop`
  - 88 files inspected
  - No offenses
- New test coverage includes:
  - `test_tombstone.rb` — 20 tests
  - `test_routing_tokens_passthrough.rb` — 32 tests
  - `test_fiber_concurrency.rb`
  - Wire-level formatter, processor, and connection-manager unit tests
- Wire formats were verified byte-for-byte against the #72 and #73 implementations.

## Review notes

- The linked compare view represents the entire review surface; there is no meaningful commit-by-commit history to review.
- Changes under `lib/` are additive. For callers that do not use the new options, `get`, `gat`, and `set` behavior remains byte-for-byte identical to upstream.
- Potential follow-up contributions to upstream, tracked separately:
  - `:miss`
  - `return_ttl_remaining:`
  - `get_multi_with_metadata`
  - `drop_value` and `tombstone_ttl` support for `meta_delete`

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -3,6 +3,7 @@ name: Benchmarks
 on:
   push:
     branches: [main]
+  pull_request:
   workflow_dispatch:
 
 jobs:
@@ -12,10 +13,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v7
-    - name: Install Memcached 1.6.23
+    - name: Install Memcached 1.6.41
       working-directory: scripts
       env:
-        MEMCACHED_VERSION: 1.6.23
+        MEMCACHED_VERSION: 1.6.41
       run: |
         chmod +x ./install_memcached.sh
         ./install_memcached.sh
--- a/.github/workflows/profile.yml
+++ b/.github/workflows/profile.yml
@@ -12,10 +12,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v7
-    - name: Install Memcached 1.6.23
+    - name: Install Memcached 1.6.41
       working-directory: scripts
       env:
-        MEMCACHED_VERSION: 1.6.23
+        MEMCACHED_VERSION: 1.6.41
       run: |
         chmod +x ./install_memcached.sh
         ./install_memcached.sh
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,7 +20,7 @@ jobs:
           - '3.3'
           - jruby-10
           - truffleruby
-        memcached-version: ['1.6.40']
+        memcached-version: ['1.6.41']
 
     name: "Ruby ${{ matrix.ruby-version }} / Memcached ${{ matrix.memcached-version }}"
     steps:
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ profile.html
 ## Environment normalisation:
 /.bundle/
 /lib/bundler/man/
+dev.yml
 
 # for a library or gem, you might want to ignore these files since the code is
 # intended to run in multiple environments; otherwise, check them in:
--- a/.standard.yml
+++ b/.standard.yml
@@ -1,6 +1,6 @@
 fix: false               # default: false
 parallel: true          # default: false
-ruby_version: 3.1       # default: RUBY_VERSION
+ruby_version: 3.3.0     # default: RUBY_VERSION
 default_ignores: false  # default: true
 
 ignore:                 # default: []
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,46 @@
 Dalli Changelog
 =====================
 
+Shopify fork (on top of upstream 5.0.5)
+==========
+
+Features:
+
+- Add opaque routing tokens `:p_token` / `:l_token` as per-request options on all
+  operations (get, gat, get_cas, get_with_metadata, fetch_with_lock, set, add,
+  replace, append, prepend, delete, incr, decr, and all multi/pipelined paths).
+  Tokens are appended to the wire as `P<token>` / `L<token>` for consumption by
+  intermediate proxies. CRLF/NUL bytes are rejected to prevent wire-protocol
+  injection; empty tokens are treated as no-ops. (Shopify#70, Shopify#72)
+- Add tombstone (mark-stale) support to `delete` / `delete_cas` / `delete_multi`
+  via the `:invalidate`, `:tombstone_ttl` and `:drop_value` request options
+  (meta-protocol `I`, `T` and `x` flags on `md`). A tombstoned item lives in a
+  stale window so concurrent readers can tell a racing repopulate apart from a
+  true miss. (Shopify#73)
+- `get_with_metadata` now returns an explicit `:miss` marker, accepts routing
+  tokens, and supports `return_ttl_remaining:` to fetch the item's remaining TTL
+  (`:ttl_remaining`, `-1` when the item has no expiry). (Shopify#73)
+- Add `get_multi_with_metadata` for stale-aware bulk reads. Returns a metadata
+  Hash (`:value`, `:cas`, `:stale`, `:miss`) for every requested key, including
+  misses, so callers can distinguish a true miss from a tombstone per key.
+  (Shopify#73)
+
+Bug Fixes:
+
+- Ensure fixed-length response reads consume exactly the requested number of
+  bytes or fail. A peer closing the connection mid-response now raises and
+  closes the dirty socket for a retry instead of surfacing a truncated value.
+  (Shopify#77)
+- Tear down the connection when a non-StandardError (e.g. `Async::Stop`,
+  `Thread#kill`) escapes a request or interrupts `ConnectionManager#close`, so
+  a half-used connection is never returned to the pool. (Shopify#68, Shopify#69,
+  Shopify#71)
+
+Performance:
+
+- Quiet-mode `ms` requests no longer include the cas-return flag, saving two
+  bytes per pipelined set.
+
 Unreleased
 ==========
 
--- a/bin/benchmark
+++ b/bin/benchmark
@@ -248,10 +248,9 @@ if %w[all set_multi].include?(bench_target)
         end
       end
     end
-    # TODO: uncomment this once we add PR adding set_multi
-    # x.report('multi client set_multi 100') do
-    #   multi_client.set_multi(pairs, 3600, raw: true)
-    # end
+    x.report('multi client set_multi 100') do
+      multi_client.set_multi(pairs, 3600, raw: true)
+    end
     x.report('write 100 keys rawsock') do
       count = pairs.length
       tail = ''
@@ -267,7 +266,16 @@ if %w[all set_multi].include?(bench_target)
       sock.flush
       sock.gets(TERMINATOR) # clear the buffer
     end
-    # x.report('write_mutli 100 keys') { client.set_multi(pairs, 3600, raw: true) }
+    x.report('write_multi 100 keys') { client.set_multi(pairs, 3600, raw: true) }
+    x.compare!
+  end
+end
+
+if %w[all delete_multi].include?(bench_target)
+  Benchmark.ips do |x|
+    x.config(warmup: bench_warmup, time: bench_time, suite: suite)
+    x.report('delete_multi 100 keys') { client.delete_multi(pairs.keys) }
+    x.report('delete 100 keys') { pairs.each_key { |key| client.delete(key) } }
     x.compare!
   end
 end
--- a/lib/dalli/client.rb
+++ b/lib/dalli/client.rb
@@ -75,8 +75,8 @@ module Dalli
     # Gat (get and touch) fetch an item and simultaneously update its expiration time.
     #
     # If a value is not found, then +nil+ is returned.
-    def gat(key, ttl = nil)
-      perform(:gat, key, ttl_or_default(ttl))
+    def gat(key, ttl = nil, req_options = nil)
+      perform(:gat, key, ttl_or_default(ttl), req_options)
     end
 
     ##
@@ -91,8 +91,8 @@ module Dalli
     ##
     # Get the value and CAS ID associated with the key.  If a block is provided,
     # value and CAS will be passed to the block.
-    def get_cas(key)
-      (value, cas) = perform(:cas, key)
+    def get_cas(key, req_options = nil)
+      (value, cas) = perform(:cas, key, req_options)
       return [value, cas] unless block_given?
 
       yield value, cas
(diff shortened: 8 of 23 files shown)


petergoldstein:
@drinkbeer — thank you for putting this together. Rebuilding the fork on top of 5.0.5 and reimplementing each change in the upstream 5.x idiom is a lot of careful work, and having it laid out section by section with the rationale for what was dropped made it straightforward to review and pull apart.

Here's how I'm planning to incorporate it.

**Sections 4 and 5 are done.** Rather than merge this branch wholesale, I've been extracting the self-contained pieces into small, individually reviewable PRs against `main`:

- #1131 — skip the cas-return flag on quiet `meta_set` (§5)
- #1132 — `set_multi` benchmark reports and a `delete_multi` target (§5)
- #1133 — CI on memcached 1.6.41, benchmarks on PRs (§5)
- #1135 — exact-length reads / truncated value fix (§4)
- #1136 — connection teardown on non-`StandardError` aborts (§4)

All five are merged, and they're rolled into the 5.0.6 patch release (prep PR: #1138). Each one carries a mutation-checked regression test — for §4 in particular, I verified that reverting each fix individually fails the corresponding test, since both bugs are the kind that only show up under conditions that are awkward to reproduce.

**Sections 1, 2 and 3 are the goal, but they need a version plan first.** Routing tokens, tombstone deletes, and the `get_with_metadata` / `get_multi_with_metadata` work all add public API, so they can't ride a patch release. Before I commit to how they land, I need to understand two things properly:

1. **Memcached version compatibility.** The `x` and `i` flags, the `t` flag for `return_ttl_remaining`, and tombstone semantics each have a minimum memcached version, and Dalli currently declares 1.6 as its floor. I need to work out what each feature actually requires and how Dalli should behave against servers that don't support it — degrade, raise, or feature-detect.
2. **The shape of the API changes.** `get_multi_with_metadata` is new surface, `get_with_metadata` gains options, and the `:miss` key changes what an existing return value means. I want to be deliberate about that before it's public and permanent.

Once those are settled I'll put together a version plan — most likely a 5.1.0 for the additive work — and continue the same extract-and-review approach. Expect PRs referencing this one.

**One small thing:** the issue references in the description and comments are being auto-linked to this repo rather than to Shopify/dalli, so they point at unrelated issues from 2011. For example, "#77 — @ianks" renders as petergoldstein/dalli#77, which is a 2011 request for UNIX domain socket support. I believe these are all meant to be `Shopify/dalli#NN`:

| Reference | Resolves here to | Intended |
|---|---|---|
| #68, #69, #71 | 2011 issues | Shopify/dalli#68, #69, #71 — connection state cleanup on non-`StandardError` |
| #70, #72 | 2011 issues | Shopify/dalli#70, #72 — meta flags passthrough, then P/L routing tokens |
| #73 | 2011 issue | Shopify/dalli#73 — meta delete with tombstone |
| #77 | 2011 issue | Shopify/dalli#77 — truncated values on mid-response close |

No rush on fixing them, but the provenance is genuinely useful when extracting these, so it'd help to have the links resolve.

Thanks also to @ianks, @nherson and @danmayer, whose original work is in here — I'm carrying credit through to the CHANGELOG entries as each section lands.

petergoldstein:
@drinkbeer — starting on §3 (the `get_with_metadata` enhancements and `get_multi_with_metadata`), and I have one API question where I'd rather not guess at your requirements.

`get_multi_with_metadata` as written returns an entry for **every requested key**, including misses. That's the opposite of Dalli's existing bulk methods, which omit them:

```ruby
c.get_multi(%w[hit nope])      #=> {"hit" => "v"}
c.get_multi_cas(%w[hit nope])  #=> {"hit" => ["v", 2]}
```

My instinct is to follow the existing convention and omit misses, because the stale-aware use case doesn't seem to need them: a tombstoned item is a protocol-level hit (`VA` with the `X` flag), so it still comes back with `stale: true`. Only genuine misses would drop out, and those are `requested - result.keys` away.

But that's exactly the kind of reasoning that's wrong when it collides with how the code is actually used. **Does anything on your side depend on misses being present in the bulk result?** Something shaped like

```ruby
result.each { |k, v| v[:miss] ? refetch(k) : use(v) }
```

would break silently under my version rather than loudly, which is the outcome I want to avoid.

If you do need every key, that's a good enough reason to keep your shape — I'd just want the asymmetry with `get_multi` documented on the method so it doesn't read as an oversight later.

For the single-key API I'm keeping your `:miss` key exactly as you wrote it, always present rather than only on misses, since `get_with_metadata` already returns a fixed set of keys (`won_recache`, `stale`, `lost_recache` are always-present booleans) and a conditionally-present key would be the odd one out there.

No rush — I'll build against "omit misses" for now since it's the smaller change to reverse, and I'll flag it clearly in the PR either way.

drinkbeer:
Thank you so much for taking the time to adopt Shopify’s changes upstream. Reducing the divergence between Shopify’s Dalli fork and the public version of Dalli will be very valuable to us, and this work will make it much easier for Shopify to upgrade to Dalli 5.x.

I audited Shopify’s internal cache usage. You can safely omit true misses from `get_multi_with_metadata`. Our internal Rails cache adapter iterates over all requested names and converts an absent Dalli result into a Rails-level miss. We also have tests covering Dalli returning both `{}` and `nil`.

We do have Rails-layer consumers shaped like your example, but they consume `Rails.cache.read_multi(..., with_status: true)`. That adapter synthesizes the missing entries before returning the result to those callers. Please continue returning stale or tombstoned `VA` responses; only genuine misses should be omitted.

There is one other API design point I would like to discuss: the return type of `get_with_metadata` and `get_multi_with_metadata`.

Upstream Dalli currently returns a metadata `Hash` from `get_with_metadata`, and the proposed multi-key API returns a `Hash` mapping each cache key to another metadata `Hash`. Shopify’s stale-aware API instead uses an immutable `Dalli::CacheResult` object with accessors and predicates such as:

```ruby
result.value
result.hit?
result.miss?
result.stale?
```

For the Dalli 5 metadata API, we could extend that concept—either as `CacheResult` or a more specific type such as `MetadataResult`—to expose `cas` and the optional metadata fields as typed attributes.

A result object has several advantages over a loosely structured hash:

- **A clearer and more discoverable API.** Callers use methods such as `miss?` and `stale?` rather than knowing and repeatedly interpreting internal hash keys.
- **Stronger typing.** Attributes can have explicit types—for example, `cas` can be constrained to `Integer` or `nil`, rather than leaving every hash value effectively untyped.
- **Unambiguous `nil` handling.** A cached `nil`, a genuine miss, and a dropped-value tombstone can be represented as distinct states without callers inferring state from `result[:value]`.
- **Enforced invariants.** Construction can reject or normalize impossible combinations such as a result being both a true miss and a stale hit.
- **Immutability.** Freezing the result prevents callers from accidentally mutating metadata returned by the client.
- **Encapsulation of protocol details.** The object can translate wire-level flags into a stable public interface, allowing internal representation changes without requiring callers to understand protocol tokens.
- **Safer evolution.** New metadata can be added through accessors and predicates while preserving existing caller behavior.

For the multi-key API, I would still expect the outer result to be something like `Hash<Key, MetadataResult>`. Ruby hashes preserve insertion order, but pipelined or multi-server response assembly does not necessarily preserve the original request order. If preserving request order is a useful API guarantee, Dalli could explicitly rebuild the returned hash in requested-key order. That ordering guarantee is separate from whether each value is represented by a hash or a result object.

I recognize that changing the existing return type of `get_with_metadata` would be a compatibility concern now that it is public API. If replacing the hash is too disruptive, possible alternatives could be a new method, an opt-in return type, or a change reserved for a future major version. I would be interested in your thoughts on whether a structured result type would be a good long-term direction for these metadata APIs.

petergoldstein:
I do like the idea of a structured result type, but I think we'd need to wait for a major release to implement that change to the return types.

In the interim we could add an adapter method along with the structured return type that library users could employ in the interim.  Not ideal from a performance perspective but it would allow us to validate the return type with a larger audience

petergoldstein:
@drinkbeer #1154 is the last unmerged piece of this draft PR.  My intention is to merge that, potentially add the adapters I suggested above for return types, and then release a 5.1.  Update for the return types would wait until a 6.0.  How does that sound?

If there are any other Shopify changes from the 3.2 fork you'd like me to consider before cutting 5.1, please let me know.

petergoldstein:
@drinkbeer — didn't want to wait on a reply before making this concrete: #1156 is a draft implementing the adapter approach discussed above. `Dalli::CacheResult` wraps `get_with_metadata`/`get_multi_with_metadata`'s existing Hash output via two new opt-in methods (`get_with_metadata_result` / `get_multi_with_metadata_result`); the existing methods and their Hash return values are untouched.

Two things left open there, both from this thread specifically:
- Naming — went with `CacheResult` to match the name your fork already uses, but `MetadataResult` is still on the table if you'd rather.
- The multi-key ordering question you raised — `get_multi_with_metadata_result` currently just inherits `get_multi_with_metadata`'s existing key order rather than rebuilding it into requested order.

Feedback welcome whenever you get to it — not blocking anything on my end.

petergoldstein:
@drinkbeer — quick update. Sections 1-3 (routing tokens, tombstone deletes, `get_with_metadata` / `get_multi_with_metadata`) are done and out: **5.1.0 has been released**, incorporating that work along with the bug fixes that landed since 5.0.6.

Next up is #1156, which adds `Dalli::CacheResult` as an interim adapter for `get_with_metadata` — planning to fold that into a 5.1.1. Feedback on the approach there is welcome before I merge it.

petergoldstein:
@drinkbeer Any comments on the above?  Should I close this PR now that everything has been incorporated?

PR #1144, opened 2026-08-02
Add get_multi_with_metadata

petergoldstein:
Second half of §3 of #1130, completing it. Builds on #1143.

Stale-aware bulk read: `{ key => { value:, cas:, stale:, miss: } }`, so a caller can tell a tombstoned item from a key that is simply gone.

```ruby
client.get_multi_with_metadata('a', 'tombstoned', 'absent')
# => { 'a'          => { value: 'v',        cas: 12, stale: false, miss: false },
#      'tombstoned' => { value: 'original', cas: 14, stale: true,  miss: false } }
```

## One deliberate deviation from #1130

**Keys that were not found are omitted**, rather than returned with `miss: true`.

`get_multi` and `get_multi_cas` both omit misses. Including them would make this the only bulk method whose result contains keys that do not exist — a caller iterating results would get entries for things that were never there.

**Nothing is lost.** A tombstoned item is a *hit* at the protocol level (`VA` with the `X` flag), so it is still returned with `stale: true` and its value. Only genuine misses drop out, and those are `requested - result.keys`. I verified this rather than assuming it — tombstoning a key with `md <key> I T30` and reading it back:

```
tomb     stale=true  miss=false value="original"
fresh    stale=false miss=false value="still here"
absent present in result? false
```

Entries keep `miss: false` even though it is constant here, so the entry shape matches `get_with_metadata` and stays stable if this is reversed — reversing then only *adds* entries rather than changing existing ones.

I asked @drinkbeer on #1130 whether anything on their side depends on the every-key shape, and built the smaller-to-reverse version while that is open. If they need it, mutation 3 below is the one-line change.

## Both code paths

`get_multi_with_metadata` routes to the single-server fast path or the pipelined getter, exactly as `get_multi` does. Both are covered: `memcached_persistent` yields a client with a two-entry ring (the pipelined path), and `single_server_client` exercises the fast path.

The pipelined variant issues one request per server and reads its full response before moving on, rather than interleaving — the metadata path has no interleaving support, and the stale-aware callers it serves fetch far smaller batches than `get_multi` does. Worth a reviewer's eye if you expect large batches through it.

## Verified against a live server

Namespace (keys returned unprefixed), raw mode (`skip_flags` path), base64-encoded keys (spaces and Unicode), non-string values, multi-server + namespace combined, empty input, and all-miss input.

## Tests

10 integration tests plus 2 formatter unit tests. Mutation-checked:

| Mutation | Failures |
|---|---|
| Drop `return_cas` from the bulk request | 2 |
| Always report `stale` as false | 1 |
| Include misses (#1130's shape) | 6 |
| Drop the `:miss` key from entries | 1 |

Full suite green (630 runs), `rubocop` clean.

The tombstone test creates its stale item at the wire level, since the client-side API for it lands in §2 — but the read path has to handle tombstones correctly now, as that distinction is the reason this method exists.


drinkbeer (commented):
The PR overall looks good to me. I just have one comment regarding the return type of `get_with_metadata` and `get_multi_with_metadata` APIs, and would like to discuss with you. See: https://github.com/petergoldstein/dalli/pull/1130#issuecomment-5200154539

petergoldstein:
@drinkbeer — following up from #1130.

Agreed this applies equally to `get_multi_with_metadata`, not just the single-key method. My plan is the same one I laid out there: the Hash return stays as-is for now, and a structured `CacheResult`/`MetadataResult` type (plus an interim adapter method, so it can be validated without breaking the current API) is major-version work. I've opened #1151 to track that so it has a home separate from this PR.

On the ordering point specifically — you're right that a multi-server fetch doesn't preserve request order in the result Hash. That's consistent with `get_multi`'s existing behavior (servers are drained via `IO.select`, not in request order), not something new here, but it was worth documenting rather than leaving implicit. Pushed 47055c2 adding a note on `get_multi_with_metadata`'s docstring.

Given the above, I'd like to merge this PR as-is and continue the return-type discussion on #1151. Let me know if that works for you.

PR #1147, opened 2026-08-03
Add opaque routing tokens (p_token / l_token) to single-key operations

petergoldstein:
§1 of #1130, single-key half.

```ruby
dc.get('key', p_token: 'pod1', l_token: 'zone2')
dc.set('key', 'value', ttl, p_token: 'pod1')
```

## What these are

Per the meta protocol spec (`protocol.txt`): *"All commands accept tokens 'P' and 'L' which are completely ignored. The arguments to 'P' and 'L' can be used as hints or path specifications to a proxy or router inbetween a client and a memcached daemon."*

I researched this before building it (see the discussion on #1130): neither vanilla memcached's built-in proxy nor mcrouter currently reads these tokens — routing in both is static config, not wire-token-driven. This is a documented protocol extension point, not tied to a specific implementation. Confirmed by bisecting memcached's source: `P`/`L` are parsed as no-ops alongside the `O` (opaque) flag since 1.6.14, well under the 1.6.27 floor from #1140.

## Scope: single-key only, and why

This touches nearly every operation, and three areas are already being modified by other open PRs adding their own new keyword parameters to the same methods. Building on those methods now would race them for the same lines:

| Deferred | Why |
|---|---|
| `get_multi` / `get_multi_cas` / `set_multi` | Not blocked, but sizable enough to be its own PR — coming next |
| `delete` / `delete_cas` / `delete_multi` / `meta_delete` | #1145 and #1146 are already adding keyword parameters to these exact methods |
| `get_multi_with_metadata` | Doesn't exist on `main` yet — defined by #1144 |

Everything else is here: `get`, `gat`, `get_cas`, `get_with_metadata`, `set`/`add`/`replace`/`set_cas`/`replace_cas`, `append`/`prepend`, `incr`/`decr`, `cas`/`cas!`, `fetch_with_lock`.

## A connection-closing bug, caught by testing against a live connection

Validating in `RequestFormatter` alone — where #1130 puts its check — means a bad token's `ArgumentError` unwinds through `Protocol::Base#request`, which logs it as an unexpected exception **and closes the connection**. I found this by testing directly against a live connection object, not by inspection:

```ruby
mgr = client.send(:ring).servers.first.instance_variable_get(:@connection_manager)
client.get('k', p_token: "bad\r\n") rescue nil
mgr.connected?  # => false, before this fix
```

**#1130's own regression test doesn't catch this.** Its equivalent test only checks that a *later* call on the same client still succeeds — which also passes when the connection silently closed and reconnected underneath it (I verified this is exactly what happens: reconnection makes the symptom invisible to a behavioral check). This PR's version asserts on `connected?` directly, and I confirmed it fails without the fix by mutation-testing it.

The fix: validate in `Client`, before `perform` is called — same pattern used for #1145's tombstone options. Added at all 16 affected entry points (some share validation via internal delegation: `set`→`set_cas`, `replace`→`replace_cas`, `cas`/`cas!`→`cas_core`).

## A gap fixed along the way

`cas_core` already accepted `req_options` but only passed it to the write half (the `set` after the CAS check) — never to the initial read. Fixed so routing tokens (and anything else in `req_options`) reach both requests `cas`/`cas!` issue.

## A deliberate extension beyond #1130's own scope

`fetch_with_lock`'s initial vivify/recache read builds its own options hash (`{ vivify_ttl:, recache_ttl: }`) and #1130 does not thread `req_options` into it — only into the fallback `set` on cache-miss regeneration. A caller passing `req_options` to `fetch_with_lock` would reasonably expect it applied to every request the call issues, not silently dropped from the first one. Fixed here; called out explicitly in case you'd rather match #1130 exactly.

## Verified beyond "it doesn't raise"

memcached ignores unknown-but-valid flags, so an operation succeeding with routing tokens attached doesn't prove they were transmitted — a formatter bug that silently dropped `p_token` would pass that check too. I wrote a wire-capture test using a fake TCP server that records raw request lines, confirming `P`/`L` actually appear on the wire for `get`, `set`, and `incr`:

```
mg k v f Ppod1 Lzone2
ms k 11 c F1 T0 MS PpodA
ma k v D1 MI LzoneX
```

## Tests

17 integration tests, 6 formatter unit tests. Mutation-checked:

| Mutation | Failures |
|---|---|
| `routing_tokens` always returns empty (tokens silently dropped) | 1 |
| Remove client-side validation on `get` | multiple |
| Remove client-side validation on `incr` (a later entry in the loop, to confirm ordering doesn't hide gaps) | 1, correctly naming `incr` |
| Drop NUL from the formatter's forbidden-byte regex | caught independently by the formatter-level unit test (client-side validation is a separate, duplicated check — defense in depth, confirmed both layers are independently real) |

Full suite green (642 runs), `rubocop` clean.


drinkbeer (approved):

petergoldstein:
Merged `main` in (`e362f45`) to resolve the conflict that came in via #1144 and #1145 — both touched `lib/dalli/protocol/base.rb`. The conflict was two independent private helpers added side by side (`routing_token_kwargs` here, `tombstone_kwargs` from #1145); kept both, no call sites overlap. Full suite (688 runs) and rubocop are clean locally.

Copilot's three review comments were already addressed in `cef8387` before this merge — the empty-string-vs-empty-object validation gap in both `RequestFormatter#routing_tokens` and `Client#validate_routing_token!`, and the `fetch_with_lock_request` merge order so `lock_ttl`/`recache_threshold` always win over a colliding `req_options` key. No further changes needed there.

PR #1152, opened 2026-08-09
Add CHANGELOG entries for #1140, #1143, #1144, #1145, #1147, #1153

petergoldstein:
CHANGELOG entries for what's landed on `main` from #1130 since the 5.0.6 release, plus the memcached floor bump that was groundwork for it (#1140 explicitly deferred its entry to this batch, matching the practice from #1137).

Covers:
- #1140 — raise the documented memcached floor to 1.6.27
- #1143 — `:miss` and `:return_ttl_remaining` on `get_with_metadata`
- #1144 — `get_multi_with_metadata`
- #1145 — tombstone deletes on `delete` / `delete_cas`
- #1147 — opaque routing tokens (`:p_token` / `:l_token`) on single-key operations

Deliberately not included: #1146 ("Add tombstone options to delete_multi"). Its two commits (`f85aee9`, `f0c0397`) only exist on `feat/tombstone-delete-multi`, merged into `feat/tombstone-deletes` — which had already been merged into `main` as #1145 a few seconds earlier. That branch was never itself merged into `main`, so despite showing as "Merged" on GitHub, none of #1146's code is actually on `main` — `delete_multi` still has no `req_options` parameter. Flagging this separately; happy to open a PR bringing those two commits onto current `main` if you want that recovered.


petergoldstein:
Rebased onto latest `main` (now includes #1153) and updated the `delete` entry to cover `delete_multi`'s tombstone support directly, rather than the "not supported yet" caveat this PR originally had — #1146's changes reached `main` via #1153 after being recovered from a stranded branch.

Issue #930, opened 2022-09-08
Performance loss when upgrading from Dalli 2.7.10 to 3.2.2


matthewzinke:
We attempted to upgrade our Dalli gem from 2.7.10 to 3.2.2 recently, and noticed a performance loss associated with the upgrade. Out memcached latency went up by just under 10% and our sidekiq processes started spending 2-3x as much time on memcached. We suspect this has something to do with the change from send_multiget to pipelined_get. 

We are wondering if anyone has also seen performance loss associated with the upgrade from 2 to 3, and/or if there are any configuration recommendations to preserve performance characteristics.

jasonpenny:
`send_multiget` looks like it was removed in https://github.com/petergoldstein/dalli/pull/852/files?diff=unified&w=1#diff-a326ee7acf56937d92cdeeb4b920547a89162170aab2ad32391200ad1b4c3638L457

was this removed just to simplify the code to always use `PipelinedGetter` @petergoldstein ?

petergoldstein:
@jasonpenny The `send_multiget` method was simply renamed to `pipelined_get` in that commit to get some naming consistency. The latter method is still there, and is now invoked by the `PipelinedGetter`.

It is certainly possible that there's a performance loss here (possibly in the `ResponseBuffer`), but I'd need to get some more detail to start debugging.

jasonpenny:
ah, I see.

I see a slowdown with a benchmark using `ActiveSupport::Cache::MemCacheStore`:

```console
root@d6ba62960da5:/# DALLI_VERSION=2.7.10 ruby bench.rb
                                            user     system      total        real
2.7.10 read_multi                       8.122310   2.830021  10.952331 ( 12.855987)
root@d6ba62960da5:/# DALLI_VERSION=2.7.10 ruby bench.rb
                                            user     system      total        real
2.7.10 read_multi                       8.037037   2.876596  10.913633 ( 12.790800)

root@d6ba62960da5:/# DALLI_VERSION=3.2.2 ruby bench.rb
                                            user     system      total        real
3.2.2 read_multi                        9.198624   2.890378  12.089002 ( 13.673892)
root@d6ba62960da5:/# DALLI_VERSION=3.2.2 ruby bench.rb
                                            user     system      total        real
3.2.2 read_multi                        9.372887   2.795796  12.168683 ( 13.765607)
```

<details>
<summary>benchmark setup in a docker container</summary
<p>

```bash
[...]

PR #1160, opened 2026-09-23
Reduce Ruby overhead on the single-key get path

radixdev:
## Summary

This PR reduces Dalli's own Ruby work on the single-key `get` path by about 28%, with no change in behavior. It addresses the remaining single-`get` regression noted in https://github.com/petergoldstein/dalli/issues/930#issuecomment-4462901142.

## Changes

- **Request building:** a plain `get` builds its request with one string interpolation (new `RequestFormatter.plain_meta_get`). It no longer goes through `meta_get`'s keyword arguments and step-by-step appends. When `get` is called with `options = nil`, it also skips option handling. The bytes sent are identical, and there is a test that compares them with `meta_get`.
- **Response parsing:** a `VA <size> f<flags>` hit line is parsed in place instead of being split into tokens and scanned for the flags token. Any other reply (`EN`, `HD`, `SERVER_ERROR`, others, and a malformed `VA` line) takes the same path and gives the same result or error as before.
- **Key check:** `KeyRegularizer.required?` now uses `/[\x00-\x20\x7F]/`. This runs only after `ascii_only?`, and over ASCII it matches exactly the same bytes as `[\p{Cntrl}\s]`. A test checks all 128 ASCII bytes against the old pattern. It is about 5× faster to match (0.22µs → 0.04µs per key), and it applies to every operation that sends a key.
- **Delegation:** `connected?` and `flushed_write` are now plain methods instead of `Forwardable` delegators, because both run on every request.

## Benchmarks

Environment: Ruby 3.4.10 and memcached 1.6.38 in Docker on Apple Silicon, with one server over loopback.

**Ruby time per `get`.** For this test the socket is replaced with an in-memory fake that replays real memcached replies, so network time does not hide the difference. Each number is the median of 7 rounds, and it was the same across two separate runs.

| | 5.1.0 | this PR | change |
|---|---:|---:|---:|
| hit | 3.45µs | 2.49µs | −28% |
| miss | 2.36µs | 1.70µs | −28% |
| hit, with an options hash (as ActiveSupport's `MemCacheStore` passes) | 3.58µs | 2.68µs | −25% |
| allocations per hit | 23 | 16 | −7 |

**With a real memcached.** The two builds alternate for 16 rounds, and each round's pair is compared:

| | paired median difference | this PR faster in |
|---|---:|---:|
| hit | −2.4µs (−6%) | 15 of 16 rounds |
| miss | −0.6µs (−2%) | 12 of 16 rounds |

In this Docker setup, the gap between 2.7.11 and 5.1.0 on single `get` was within run-to-run noise, so I could not measure the ~12% from the comment above. This PR removes about 1µs of Ruby work per call. That share of the total is larger on native setups with less loopback overhead, like the one in that comment.

Things I tried that did **not** help: other socket read strategies (`IO#gets` with `IO#timeout`, `read_nonblock`, `wait_readable`) were all within 0.3µs of each other. 5.1.0 already makes the same syscalls per `get` as 2.7.11 (one write, one read).

<details>
<summary>Fake-socket harness used for the Ruby-only numbers</summary>

```ruby
require 'dalli'; require 'socket'

class FakeSock # replays canned meta responses; no syscalls
  def initialize(table) = (@table = table; @buf = +''; @pos = 0)
  def write(s) = (@buf << @table.fetch(s[/\Amg (\S+)/, 1]); s.bytesize)
  def flush = self
  def gets(sep) = (i = @buf.index(sep, @pos); line = @buf.byteslice(@pos, i + sep.bytesize - @pos); @pos = i + sep.bytesize; compact; line)
  def read(n) = (s = @buf.byteslice(@pos, n); @pos += n; compact; s)
  def compact = ((@buf.clear; @pos = 0) if @pos >= @buf.bytesize)
  def close; end
  def closed? = false
end

c = Dalli::Client.new('127.0.0.1:11211', compress: false)
c.set('hit', 'x' * 100); c.get('hit')
raw = TCPSocket.new('127.0.0.1', 11211)
table = %w[hit miss].to_h do |k|
  raw.write("mg #{k} v f\r\n"); line = raw.gets("\r\n")
  [k, line + (line.start_with?('VA') ? raw.read(line.split[1].to_i + 2) : '')]
end
cm = c.instance_variable_get(:@ring).servers.first.instance_variable_get(:@connection_manager)
cm.instance_variable_set(:@sock, FakeSock.new(table))
# ...then time c.get('hit') / c.get('miss') in a loop, and check that cm.sock is still the FakeSock afterwards
```
</details>

## Testing

- `bundle exec rake`: 744 runs, 0 failures, 0 errors. This includes 6 new tests: for `plain_meta_get`, for the key check over all ASCII bytes, and for hit lines with flags in other positions, a missing flags token, and `SERVER_ERROR`/unexpected replies.
- The new response-parsing and key-check tests also pass against unchanged 5.1.0, which confirms the behavior is the same.
- `bundle exec rubocop` reports no new offenses. With the current rubocop release (1.91.0), `main` already has 10 offenses, mostly `Style/DirectiveScope`. Two of them are in `base.rb`, but not on lines this PR changes. I left them alone to keep this diff focused, so the rubocop check may fail for that reason.

## Notes

- #1130 also edits `ResponseProcessor#error_on_unexpected!`. This PR moves the two `raise` lines from that method into a new `raise_unexpected!` helper, so whichever PR lands second may need a small rebase.
- This PR was generated with the assistance of Claude Code (Anthropic). The benchmark and test results above come from real runs.


radixdev:
Hi @petergoldstein ,

I reviewed the contribution guidelines and I hope you can take a look 🙏🏽 .

Thanks,
Some dude at Braze

petergoldstein:
Thanks @radixdev !

PR #1077, opened 2026-02-20
Add single-server fast path for get_multi

danmayer:
When only one memcached server is configured (common for many deployments when using mcrouter, memcached build in proxy, envoy, etc), bypass the PipelinedGetter machinery (IO.select, response buffering, server grouping) and issue all quiet meta get requests inline followed by a noop terminator.

__Benchmarks show ~1.5x improvement at 10 keys and ~1.75x at 100-500 keys compared to the PipelinedGetter path.__

This upstreams work available in the Shopify fork that originally was added in this PR: https://github.com/Shopify/dalli/pulls?page=3&q=is%3Apr+is%3Aclosed

* disclosure: I used Claude code to extract the single server optimized multi_get from our Shopify fork
* If you are interested in accepting these patches, I can either make new PRs or extend this PR to include multi_set and multi_delete single server optimizations that we have, which show similar improvements. Let me know your thoughts @petergoldstein 
* While reducing allocations and improvements for the ring is also very valuable, this optimized approach has helped to significantly reduce latency under high load.
* I included the benchmark script as it is useful for testing other optimization PRs, but I am happy to remove that or just dump it in a gist

  | Benchmark | Baseline (i/s) | Fast Path (i/s) | Baseline (μs/i) | Fast Path (μs/i) | Speedup |
  |---|---|---|---|---|---|
  | `get_multi(10)` | 6,257 | 9,347 | 159.8 | 107.0 | **1.49x** |
  | `get_multi(100)` | 1,039 | 1,795 | 962.9 | 557.1 | **1.73x** |
  | `get_multi(500)` | 235 | 410 | 4,264 | 2,437 | **1.75x** |

PR #1161, opened 2026-09-23
Speed up multi-server get_multi

radixdev:
## Summary

This PR speeds up `get_multi` on multi-server rings by 20–31% over loopback (4 servers). It brings multi-server `get_multi` level with Dalli 2.7.11. The single-server fast path from #1077 does not help multi-server rings, so this is the case that https://github.com/petergoldstein/dalli/issues/930 left open: "Multi-server deployments wouldn't benefit from the fast path and I haven't benchmarked that case."

This PR is independent of #1160. The two PRs touch nearby lines in 4 files, so whichever lands second will need a small rebase.

## Why multi-server `get_multi` was slow

These are profiled on 4 servers with 100 keys:

1. **Requests are sent late.** `PipelinedGetter#setup_requests` built and buffered every server's queries before it wrote any server's terminating `mn`. So memcached started work only after all the client-side work was done, and its response time never overlapped with anything. 2.7.11 sends each server's request as soon as it's built. For small batches this was the whole gap: a 3-key `get_multi` makes *fewer* syscalls than 2.7.11 does (10 against 13), yet it was about 20% slower. A CPU profile doesn't show this, because the extra time is spent waiting.
2. **Reply parsing** took about 38% of CPU. `getk_response_from_buffer` split each header, then scanned the tokens once each for `s`, `f`, `c` and `k`, and again for `b`.
3. **Routing keys** took about 15% of CPU. Each key called `alive?` twice (once in `server_from_continuum` and again in `server_for_key`), and each call takes the server's lock. The ring's `bsearch_index` also called `Entry#value` at every step.
4. **Building requests** took about 10% of CPU. Each key went through `meta_get`'s 14 keyword arguments. Plain `get_multi` also requested a CAS value for every key and then threw it away.

## Changes

- **`PipelinedGetter#setup_requests`** sends each server's queries and `mn` before it builds the next server's. It is still one write per server. The per-server error handling stays the same:
  - `make_getkq_request` (per server) still re-raises network errors and swallows a plain `DalliError`.
  - A network error aborts the servers that already have requests pending, then re-raises so `#process` retries.
  - A `DalliError` from the noop drops only that server.
- **`Base#pipelined_get`** builds the batch in one pass with `RequestFormatter.multi_meta_get`. A new `terminate: false` option leaves off the `mn`, which `pipeline_response_setup` still sends. A new `return_cas` argument (default `true`) lets plain `get_multi` leave off the `c` flag. `get_multi_cas` still requests it.
- **`getk_response_from_buffer`** reads the flags in one pass, keeping the old rule that the first matching token wins. On 16 cases, including partial buffers, base64 keys, missing flags and error lines, its output is identical to the old code's.
- **`Ring`:**
  - `keys_grouped_by_server` checks each server's `alive?` once per call.
  - `server_for_key` no longer repeats the `alive?` check that `server_from_continuum` already did. This also saves one lock on every single-key operation in a multi-server ring.
  - The binary search runs over plain integers. A `continuum=` writer keeps them in sync with the entries.
- **Smaller per-call costs:**
  - After `IO.select`, a socket's server is found with `Array#index`. This keeps the existing linear-scan approach, but the scan runs in C.
  - `Base#sock` is now a plain method instead of a `Forwardable` delegator.
  - `get_multi` builds trace attributes only when tracing is enabled.

## Benchmarks

Environment:
- Ruby 3.4.10 and memcached 1.6.38, in Docker on Apple Silicon.
- 4 memcached servers.
- Each number is the fastest of 6 slices across 4 rounds, with the three builds interleaved.
- memcached and Ruby are pinned to separate CPUs.

**Loopback:**

| keys | 2.7.11 | 5.1.0 | this PR | vs 5.1.0 |
|---:|---:|---:|---:|---:|
| 3 | 53.2µs | 65.2µs | 52.2µs | −20% |
| 20 | 111.5µs | 133.5µs | 98.6µs | −26% |
| 100 | 303.3µs | 443.1µs | 306.6µs | −31% |
| allocations, 100 keys | 1,413 | 2,142 | 1,536 | −28% |

**About 300µs round trip** (`tc netem`, 100µs per packet):

| keys | 2.7.11 | 5.1.0 | this PR | vs 5.1.0 |
|---:|---:|---:|---:|---:|
| 3 | 337.7µs | 346.7µs | 337.6µs | −3% |
| 20 | 377.4µs | 418.8µs | 379.3µs | −9% |
| 100 | 552.4µs | 736.0µs | 576.4µs | −22% |

Most of the 100-key gain comes from the parsing and routing changes. Sending requests earlier is what brings small batches level with 2.7.11. What's left at 100 keys over a real network (+4%) is mostly `split` and `Marshal.load` per reply.

## Testing

- `bundle exec rake`: 750 runs, 0 failures, 0 errors. That includes 12 new tests:
  - A **real 2-server ring**. The existing `get_multi` integration tests use a single server, so they take the fast path. The new tests cover the hash and block forms, misses, a namespace, keys with spaces or non-ASCII characters, and `get_multi_cas` returning real CAS values.
  - The send order in `setup_requests` and its error handling.
  - `alive?` being called once per server, and once per single-key lookup.
  - The one-pass parser: flags in any order, a base64 key, a missing CAS, and a partial body.
  - `multi_meta_get(terminate: false)`.
- One existing test changed: `test_routing_tokens_bulk.rb` asserted the exact arguments of the `:pipelined_get` dispatch. It now checks the first two, which are the keys and the threaded `req_options`, because `return_cas` is now a third argument.
- `bundle exec rubocop` reports no new offenses. `main` already has 10 offenses with rubocop 1.91.0, mostly `Style/DirectiveScope` (as noted in #1160), so the rubocop check may fail for that reason.
- `pipelined_get` takes `return_cas` as a positional argument because `Base#request` forwards only positional arguments. It carries an inline `Style/OptionalBooleanParameter` disable with that reason.

## Notes

- The only change on the wire: plain `get_multi` sends `mg <key> v f k q s` instead of `mg <key> v f c k q s`.
- This PR was generated with the assistance of Claude Code (Anthropic). The benchmark and test results above come from real runs.


petergoldstein:
Thanks @radixdev !

